PR #1, opened 2024-01-11
Add SingleAddressWallet

n8maninger:


testutil/wallet.go
```diff
@@ -0,0 +1,265 @@
...
+
+// ProcessChainApplyUpdate implements chain.Subscriber
+func (es *EphemeralWalletStore) ProcessChainApplyUpdate(cau *chain.ApplyUpdate, commit bool) error {
+	es.mu.Lock()
+	defer es.mu.Unlock()
+
+	es.uncommitted = append(es.uncommitted, cau)
+	if !commit {
+		return nil
+	}
```
lukechampine, line 77: this confuses me. Subscribers that don't write anything to persistent storage should never need to buffer updates like this.

n8maninger: I copied it mostly to show how a real implementation would need to work

go.sum
```diff
@@ -4,19 +4,25 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
...
-golang.org/x/crypto v0.0.0-20220507011949-2cf3adece122 h1:NvGWuYG8dkDHFSKksI1P9faiVJ9rayE6l0+ouWVIDs8=
-golang.org/x/crypto v0.0.0-20220507011949-2cf3adece122/go.mod h1:IxCIyHEi3zRg3s0A5j5BB6A9Jmi73HwBIUl50j+osU4=
+go.sia.tech/mux v1.2.0 h1:ofa1Us9mdymBbGMY2XH/lSpY8itFsKIo/Aq8zwe+GHU=
+go.sia.tech/mux v1.2.0/go.mod h1:Yyo6wZelOYTyvrHmJZ6aQfRoer3o4xyKQ4NmQLJrBSo=
+go.uber.org/goleak v1.2.0 h1:xqgm/S+aQvhWFTtR0XK3Jvg7z8kGV8P4X14IzwN3Eqk=
+go.uber.org/goleak v1.2.0/go.mod h1:XJYK+MuIchqpmGmUSAzotztawfKvYLUIgg7guXrwVUo=
+go.uber.org/multierr v1.11.0 h1:blXXJkSxSSfBVBlC76pxqeO+LN3aDfLQo+309xJstO0=
+go.uber.org/multierr v1.11.0/go.mod h1:20+QtiLqy0Nd6FdQB9TLXag12DsQkrbs3htMFfDN80Y=
+go.uber.org/zap v1.26.0 h1:sI7k6L95XOKS281NhVKOFCUNIvv9e0w4BF8N3u+tCRo=
+go.uber.org/zap v1.26.0/go.mod h1:dtElttAiwGvoJ/vj4IwHBS/gXsEu/pZ50mUIRWuG0so=
```
ChrisSchinnerl, line 20: Could you also update the `Syncer` to use the zap.Logger to make sure all of `coreutils` uses the same?

wallet/wallet.go
```diff
@@ -0,0 +1,522 @@
...
+		if err != nil { // no way to check for specific reorg error, so reset everything
+			sw.log.Warn("rescanning blockchain due to unknown consensus change ID")
+			// reset change ID and subscribe again
+			if err := store.ResetWallet(seedHash); err != nil {
+				sw.log.Fatal("failed to reset wallet", zap.Error(err))
+			} else if err = cm.AddSubscriber(store, types.ChainIndex{}); err != nil {
+				sw.log.Fatal("failed to reset consensus change subscription", zap.Error(err))
+			}
+		}
+	}()
```
ChrisSchinnerl, line 520: Can we remove subscribing the `store` from `NewSingleAddressWallet` and expect the user to pass in an already subscribed store which resets itself in case there is some inconsistency?

Because our store in the renter subscribes for more than just the wallet so it would be a bit awkward to have the following

```go
store := newSQLStore() // serves as object store, wallet store, host store etc.

saw := newSingleAddressWallet(store) // subscribes store to consensus

b := newBus(saw, store)
```

That would also get rid of `ResetWallet` from the interface if we just expect the caller to handle an invalid subscription when creating the store. Then it would look more like

```go
store := newSQLStore()

subscribeToConsensus(store)
if err {
  store.ResetSubscription()
}

saw := newSingleAddressWallet(store)
```

wallet/wallet.go
```diff
@@ -0,0 +1,522 @@
...
+		cm    ChainManager
+		store SingleAddressStore
+		log   *zap.Logger
+
+		mu  sync.Mutex // protects the following fields
+		tip types.ChainIndex
+		// locked is a set of siacoin output IDs locked by FundTransaction. They
+		// will be released either by calling Release for unused transactions or
+		// being confirmed in a block.
+		locked map[types.Hash256]bool
```
ChrisSchinnerl, line 112: would be nice to change this to `locked map[types.Hash256]time.Time` where the timestamp is the last time it was locked like we do in `renterd` and then pass in an argument when creating the wallet to determine for how long an output is locked before trying to spend it again.

It might be different in `hostd` but in `renterd` there is no 100% guarantee that an output is unlocked upon error in a distributed setup.

wallet/wallet.go
```diff
@@ -0,0 +1,522 @@
...
+	"go.sia.tech/core/chain"
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+	"go.uber.org/zap"
+)
+
+const (
+	// transactionDefragThreshold is the number of utxos at which the wallet
+	// will attempt to defrag itself by including small utxos in transactions.
+	transactionDefragThreshold = 30
```
ChrisSchinnerl, line 19: Might be nice to have this configurable in a shared implementation. In case `renterd` and `hostd` want to use different values.

wallet/wallet.go
```diff
@@ -0,0 +1,522 @@
...
+			continue
+		}
+		usableUTXOs = append(usableUTXOs, sce)
+	}
+
+	// sort by value, descending
+	sort.Slice(usableUTXOs, func(i, j int) bool {
+		return usableUTXOs[i].SiacoinOutput.Value.Cmp(usableUTXOs[j].SiacoinOutput.Value) > 0
+	})
+
```
ChrisSchinnerl, line 307: Could we add an argument to `FundTransaction` called `useUnconfirmed` and then check here whether it's true and if so append unconfirmed transactions (also sorted by value) to `usableUTXOs`?

This makes a day and night difference when forming contracts on a fresh renter.

wallet/wallet.go
```diff
@@ -0,0 +1,522 @@
...
+	release := func() {
+		sw.mu.Lock()
+		defer sw.mu.Unlock()
+		for _, id := range toSign {
+			delete(sw.locked, id)
+		}
+	}
+	return toSign, release, nil
+}
+
```
ChrisSchinnerl, line 372: would be nice to include our `Redistribute` method.

The original looks like this:
```go
// Redistribute returns a transaction that redistributes money in the wallet by
// selecting a minimal set of inputs to cover the creation of the requested
// outputs. It also returns a list of output IDs that need to be signed.
func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) ([]types.Transaction, []types.Hash256, error) {
	w.mu.Lock()
	defer w.mu.Unlock()

	// build map of inputs currently in the tx pool
	inPool := make(map[types.Hash256]bool)
	for _, ptxn := range pool {
		for _, in := range ptxn.SiacoinInputs {
			inPool[types.Hash256(in.ParentID)] = true
		}
	}

	// fetch unspent transaction outputs
	utxos, err := w.store.UnspentSiacoinElements(false)
	if err != nil {
		return nil, nil, err
	}

	// check whether a redistribution is necessary, adjust number of desired
	// outputs accordingly
	for _, sce := range utxos {
		inUse := w.isOutputUsed(sce.ID) || inPool[sce.ID]
		matured := cs.Index.Height >= sce.MaturityHeight
		sameValue := sce.Value.Equals(amount)
		if !inUse && matured && sameValue {
			outputs--
		}
	}
	if outputs <= 0 {
		return nil, nil, nil
	}

	// desc sort
	sort.Slice(utxos, func(i, j int) bool {
		return utxos[i].Value.Cmp(utxos[j].Value) > 0
	})

	// prepare all outputs
	var txns []types.Transaction
	var toSign []types.Hash256

	for outputs > 0 {
		var txn types.Transaction
		for i := 0; i < outputs && i < redistributeBatchSize; i++ {
			txn.SiacoinOutputs = append(txn.SiacoinOutputs, types.SiacoinOutput{
				Value:   amount,
				Address: w.Address(),
			})
		}
		outputs -= len(txn.SiacoinOutputs)

		// estimate the fees
		outputFees := feePerByte.Mul64(uint64(len(encoding.Marshal(txn.SiacoinOutputs))))
		feePerInput := feePerByte.Mul64(BytesPerInput)

		// collect outputs that cover the total amount
		var inputs []SiacoinElement
		want := amount.Mul64(uint64(len(txn.SiacoinOutputs)))
		var amtInUse, amtSameValue, amtNotMatured types.Currency
		for _, sce := range utxos {
			inUse := w.isOutputUsed(sce.ID) || inPool[sce.ID]
			matured := cs.Index.Height >= sce.MaturityHeight
			sameValue := sce.Value.Equals(amount)
			if inUse {
				amtInUse = amtInUse.Add(sce.Value)
				continue
			} else if sameValue {
				amtSameValue = amtSameValue.Add(sce.Value)
				continue
			} else if !matured {
				amtNotMatured = amtNotMatured.Add(sce.Value)
				continue
			}

			inputs = append(inputs, sce)
			fee := feePerInput.Mul64(uint64(len(inputs))).Add(outputFees)
			if SumOutputs(inputs).Cmp(want.Add(fee)) > 0 {
				break
			}
		}

		// not enough outputs found
		fee := feePerInput.Mul64(uint64(len(inputs))).Add(outputFees)
		if sumOut := SumOutputs(inputs); sumOut.Cmp(want.Add(fee)) < 0 {
			// in case of an error we need to free all inputs
			w.releaseInputs(txns...)
			return nil, nil, fmt.Errorf("%w: inputs %v < needed %v + txnFee %v (usable: %v, inUse: %v, sameValue: %v, notMatured: %v)",
				ErrInsufficientBalance, sumOut.String(), want.String(), fee.String(), sumOut.String(), amtInUse.String(), amtSameValue.String(), amtNotMatured.String())
		}

		// set the miner fee
		txn.MinerFees = []types.Currency{fee}

		// add the change output
		change := SumOutputs(inputs).Sub(want.Add(fee))
		if !change.IsZero() {
			txn.SiacoinOutputs = append(txn.SiacoinOutputs, types.SiacoinOutput{
				Value:   change,
				Address: w.addr,
			})
		}

		// add the inputs
		for _, sce := range inputs {
			txn.SiacoinInputs = append(txn.SiacoinInputs, types.SiacoinInput{
				ParentID:         types.SiacoinOutputID(sce.ID),
				UnlockConditions: StandardUnlockConditions(w.priv.PublicKey()),
			})
			toSign = append(toSign, sce.ID)
			w.lastUsed[sce.ID] = time.Now()
		}

		txns = append(txns, txn)
	}

	return txns, toSign, nil
}
```

testutil/wallet.go
lukechampine: A subscriber with a transactional DB wouldn't look like this either, though. You'd write things as usual and then commit the transaction on `mayCommit`. 🤔

n8maninger: With many transactional databases, holding transactions open for long periods is bad practice. In SQLite, when a transaction writes, all other threads are blocked until it's committed. In MySQL and Postgres, you will hit table locks. Those stores would require a buffer to cache before opening a transaction.

The best practice is to open a transaction, write everything, and commit or roll back as quickly as possible.

go.sum
n8maninger: #3

wallet/wallet.go
n8maninger: I know I've wanted this function before, but can you remind me of the benefit of this vs `FundTransaction` with ten siacoin outputs? Is it only the miner fee calculation?

ChrisSchinnerl: If you have 30 outputs of value X and you want 50, `Redistribute` can avoid touching the 30 outputs and create 20 more. 
`FundTransaction` with `n=20` might reuse the 30 outputs we already have so we need to do `n=50` which might result in anywhere from 50 to 70 outputs of the value we want.

n8maninger: Going to handle this in a follow-up

PR #3, opened 2024-01-13
Use structured log for syncer

n8maninger:


lukechampine:
You know, the original version of the Syncer took a `Logger` interface. I ended up removing it because it felt redundant. But this makes me want to put it back in. Coupling the Syncer to some third-party dependency feels sinful.

Has anyone looked at https://pkg.go.dev/log/slog yet? It was added in 1.21.

n8maninger:
I'm most familiar with Zap, but I wouldn't mind migrating everything to `slog` as long as we're consistent. I looked at `slog` when it was announced, but I preferred Zap's API. I think it allocates less too.

> Coupling the Syncer to some third-party dependency feels sinful.

"lukechampine.com/frand" 👀

lukechampine:
funny you mention that -- Go 1.22 has a new CSPRNG that will make `frand` obsolete: https://github.com/C2SP/C2SP/blob/main/chacha8rand.md

I am slightly sad about it (since I am proud of `frand`) but overall it will be nice to have a proper CSPRNG in the stdlib. And the design has a lot of overlap with `frand`, so I'll view it as a spiritual successor :P

PR #14, opened 2024-01-31
Add Announcement type and ForEachAnnouncement

ChrisSchinnerl:


chain/announcements.go
```diff
@@ -0,0 +1,99 @@
...
+			var ha V1Announcement
+			dec := types.NewBufDecoder(arb)
+			ha.DecodeFrom(dec)
+			if err := dec.Err(); err != nil {
+				continue
+			} else if ha.Specifier != types.NewSpecifier(AnnouncementSpecifier) {
+				continue
+			}
+			// verify signature
+			if !ha.VerifySignature() {
```
peterjan, line 77: `nit:` I'd turn this in an `else if` and join the validation?

chain/announcements.go
```diff
@@ -0,0 +1,99 @@
...
+			copy(pk[:], ha.PublicKey.Key)
+			fn(Announcement{
+				NetAddress: ha.NetAddress,
+				PublicKey:  pk,
+			})
+		}
+	}
+	for _, txn := range b.V2Transactions() {
+		for _, att := range txn.Attestations {
+			if att.Key != AnnouncementSpecifier {
```
peterjan, line 90: should we add something similar to `VerifySignature` here or are attestations already verified somewhere?

peterjan (commented):
LGTM 👍

chain/announcements.go
ChrisSchinnerl: `V2` attestations are already verified here

PR #23, opened 2024-02-16
Fix manager panic

chris124567:
Fix case where chain.Manager would panic when adding a v1 transaction after the v2 hardfork require height.


lukechampine:
I decided to move the check into the `Store`; I think it's a slight improvement to say "the supplement for a v2 block is defined as empty," rather than checking the height wherever we might need to compute a supplement. ‾\\\_(ツ)\_/‾

lukechampine (approved):

ChrisSchinnerl (approved):

PR #24, opened 2024-02-21
Don't use immature outputs when funding a transaction

peterjan:


wallet/wallet.go
```diff
@@ -274,6 +274,17 @@ func (sw *SingleAddressWallet) FundTransaction(txn *types.Transaction, amount ty
 		return nil, err
 	}
 
+	cs := sw.cm.TipState()
+
+	// filter out immature elements
+	filtered := elements[:0]
```
n8maninger, line 280: This should be added to the existing loop on line 311

n8maninger (changes requested):

n8maninger (approved):

ChrisSchinnerl (approved):

PR #27, opened 2024-03-06
Fix 'inUse' condition in wallet.Redistribute

peterjan:
Oversight in the `inUse` condition, `time.After` vs `time.Before` bug, should've been caught so added a regression test.

### DIFF
--- a/testutil/network.go
+++ b/testutil/network.go
@@ -22,7 +22,8 @@ func Network() (*consensus.Network, types.Block) {
 	return n, genesisBlock
 }
 
-// MineBlock mines a block with the given transactions.
+// MineBlock mines a block with the given transactions, transaction fees are
+// added to the miner payout.
 func MineBlock(cm *chain.Manager, minerAddress types.Address) types.Block {
 	var minerFees types.Currency
 	for _, txn := range cm.PoolTransactions() {
--- a/testutil/network.go
+++ b/testutil/network.go
@@ -24,12 +24,17 @@ func Network() (*consensus.Network, types.Block) {
 
 // MineBlock mines a block with the given transactions.
 func MineBlock(cm *chain.Manager, minerAddress types.Address) types.Block {
+	var minerFees types.Currency
+	for _, txn := range cm.PoolTransactions() {
+		minerFees = minerFees.Add(txn.TotalFees())
+	}
+
 	state := cm.TipState()
 	b := types.Block{
 		ParentID:     state.Index.ID,
 		Timestamp:    types.CurrentTimestamp(),
 		Transactions: cm.PoolTransactions(),
-		MinerPayouts: []types.SiacoinOutput{{Address: minerAddress, Value: state.BlockReward()}},
+		MinerPayouts: []types.SiacoinOutput{{Address: minerAddress, Value: state.BlockReward().Add(minerFees)}},
 	}
 	for b.ID().CmpWork(state.ChildTarget) < 0 {
 		b.Nonce += state.NonceFactor()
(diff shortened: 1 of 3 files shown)


peterjan:
would be nice if we could retag `coreutils` after merging this, maybe we can take the subscriber refactor along in the new release

testutil/network.go
```diff
@@ -24,12 +24,17 @@ func Network() (*consensus.Network, types.Block) {
 
 // MineBlock mines a block with the given transactions.
 func MineBlock(cm *chain.Manager, minerAddress types.Address) types.Block {
+	var minerFees types.Currency
+	for _, txn := range cm.PoolTransactions() {
```
ChrisSchinnerl, line 28: Could you update the docstring while you are at it to match this code?

wallet/wallet.go
```diff
@@ -496,7 +496,7 @@ func (sw *SingleAddressWallet) Redistribute(outputs int, amount, feePerByte type
 	// unused, matured and has the same value
 	utxos := make([]types.SiacoinElement, 0, len(elements))
 	for _, sce := range elements {
-		inUse := time.Now().After(sw.locked[sce.ID]) || inPool[sce.ID]
+		inUse := time.Now().Before(sw.locked[sce.ID]) || inPool[sce.ID]
```
ChrisSchinnerl, line 499: Is this the only spot where we compute `inUse`? Otherwise we could make that a helper

wallet/wallet_test.go
```diff
@@ -378,3 +297,157 @@ func TestWalletUnconfirmed(t *testing.T) {
...
+			return nil, nil, nil, err
+		}
+	}
+
+	return w, cm, ws, nil
+}
+
+// check balance is a helper function that compares the wallet's balance to
+// the expected values.
+func checkBalance(w *wallet.SingleAddressWallet, spendable, confirmed, immature, unconfirmed types.Currency) error {
```
ChrisSchinnerl: If this is in the same package you could make this a method on `SingleAddressWallet`. I'd also prefer the name `assertBalance`.

peterjan: Sorry, I updated the code right when you were reviewing it seems. I'm going to resolve this because I did something similar. Feel free to re-open though if you don't like the `testWallet`.

lukechampine:
> would be nice if we could retag `coreutils` after merging this, maybe we can take the subscriber refactor along in the new release

fwiw, you don't *need* a tag in order to update a dependency; can always reference a commit directly in your go.mod (though tags are still nice for human legibility)

ChrisSchinnerl:
> > would be nice if we could retag `coreutils` after merging this, maybe we can take the subscriber refactor along in the new release
> 
> fwiw, you don't _need_ a tag in order to update a dependency; can always reference a commit directly in your go.mod (though tags are still nice for human legibility)

We definitely shouldn't tag too often but I think tags have 2 advantages over commits.
1. It's easier to keep all node software on the same exact version
2. Tooling like dependabot will picks new tags up but not commits

ChrisSchinnerl (dismissed):

wallet/wallet_test.go
```diff
@@ -95,7 +169,7 @@ func TestWallet(t *testing.T) {
 	}
 
 	// try funding the transaction, expect it to fail since the outputs are immature
-	_, err = w.FundTransaction(&txn, initialReward, false)
+	_, err := w.FundTransaction(&txn, initialReward, false)
 	if err != wallet.ErrNotEnoughFunds {
```
ChrisSchinnerl, line 173: This needs to be updated now to use `errors.Is`.

ChrisSchinnerl (changes requested):

wallet/wallet_test.go
peterjan: woopsie

ChrisSchinnerl (approved):

PR #28, opened 2024-03-06
Add context to syncer.Connect

peterjan:
I had some TODOs in `renterd` for missing contexts in some of the new `coreutils`, but after closer inspection I can see why no contexts were added. Not sure if I totally agree in all cases though, for instance the chain manager never takes a context, even though the store is injected and might be a remote database or something so I think there's something to say for extending the `Store interface` and add a context to literally every method... That said, I think `syncer.Connect` is the only one that's really missing a context so I decided to suggest adding it.

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -411,6 +411,12 @@ func (s *Syncer) acceptLoop() error {
 	}
 }
 
+func (s *Syncer) isStopped() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.l == nil
+}
+
 func (s *Syncer) peerLoop(closeChan <-chan struct{}) error {
 	log := s.log.Named("peerLoop")
 	numOutbound := func() (n int) {
@@ -479,11 +485,6 @@ func (s *Syncer) peerLoop(closeChan <-chan struct{}) error {
 			return false
 		}
 	}
-	closing := func() bool {
-		s.mu.Lock()
-		defer s.mu.Unlock()
-		return s.l == nil
-	}
 	for fst := true; fst || sleep(); fst = false {
 		if numOutbound() >= s.config.MaxOutboundPeers {
 			continue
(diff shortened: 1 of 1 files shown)


syncer/syncer.go
```diff
@@ -497,9 +497,13 @@ func (s *Syncer) peerLoop(closeChan <-chan struct{}) error {
...
 				break
 			}
+
 			// NOTE: we don't bother logging failure here, since it's common and
 			// not particularly interesting or actionable
-			if _, err := s.Connect(p); err == nil {
+			//
+			// NOTE: we can safely pass the background context here since the
+			// connect timeout will be enforced
+			if _, err := s.Connect(context.Background(), p); err == nil {
```
n8maninger, line 506: Should `ctx, cancel := context.WithTimeout(context.Background(), s.config.ConnectTimeout)` be moved out of `Connect()` to give users full control over the lifecycle when calling `Connect()` directly?

lukechampine: yes

syncer/syncer.go
```diff
@@ -635,27 +641,10 @@ func (s *Syncer) Run() error {
...
 // Connect forms an outbound connection to a peer.
-func (s *Syncer) Connect(addr string) (*Peer, error) {
+func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
 	if err := s.allowConnect(addr, false); err != nil {
 		return nil, err
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), s.config.ConnectTimeout)
-	defer cancel()
-	// slightly gross polling hack so that we shutdown quickly
-	go func() {
```
ChrisSchinnerl, line 645: Why can we get rid of this now? I'd expect this to be replaced with something like

```go
go func() {
  select {
    case <-ctx.Done():
    case <-s.shutdownCtx.Done():
      cancel()
  }
}

To make sure we interrupt both on `ctx` getting closed as well as `shutdownCtx` being cancelled.

File: syncer/syncer.go [hunk 3]
```diff
@@ -411,7 +414,13 @@ func (s *Syncer) acceptLoop() error {
 	}
 }
 
-func (s *Syncer) peerLoop(closeChan <-chan struct{}) error {
+func (s *Syncer) isStopped() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.l == nil
```
ChrisSchinnerl, line 420: Considering that we have a `shutdownCtx` as a signal now I'd probably use that instead of relying on the listener being `nil`.

e.g.
```go
select {
  case <-s.shutdownCtx:
    return true
  default:
    return false
}
```

ChrisSchinnerl (changes requested):

syncer/syncer.go
peterjan: I considered that but didn't want to change it because of the slight nuance, this would return true before the listener gets closed. It's likely fine though I'll update it 👍

syncer/syncer.go
ChrisSchinnerl: I think it's only fine to drop if this method is not exposed. If someone can call it who doesn't have access to `shutdownCtx` we need to explicitly abort.

ChrisSchinnerl (approved):

PR #37, opened 2024-03-12
Expose ExplicitCoveredFields from the wallet package

peterjan:


lukechampine:
fwiw this will be obsolete in v2, but I suppose it's fine to put it here if both the renter and host are using it.

PR #38, opened 2024-03-12
Add context to every method in the PeerStore interface

peterjan:
I think every `XStore` method should have a context. We might not care all the time but someone implementing the Store interface themselves might and not accepting a context might create inconsistent code on their side (running into this in `renterd`) In the syncer I could use the new `shutdownCtx`, applied a generous 5' timeout here and there which I think is fine but open to suggestions.

### DIFF
--- a/syncer/peer.go
+++ b/syncer/peer.go
@@ -223,7 +223,7 @@ func (p *Peer) acceptRPC() (types.Specifier, *gateway.Stream, error) {
 func (s *Syncer) handleRPC(id types.Specifier, stream *gateway.Stream, origin *Peer) error {
 	switch r := gateway.ObjectForID(id).(type) {
 	case *gateway.RPCShareNodes:
-		peers, err := s.pm.Peers()
+		peers, err := s.pm.Peers(s.shutdownCtx)
 		if err != nil {
 			return fmt.Errorf("failed to fetch peers: %w", err)
 		} else if n := len(peers); n > 10 {
--- a/syncer/peer.go
+++ b/syncer/peer.go
@@ -1,6 +1,7 @@
 package syncer
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
@@ -220,10 +221,10 @@ func (p *Peer) acceptRPC() (types.Specifier, *gateway.Stream, error) {
 	return id, s, nil
 }
 
-func (s *Syncer) handleRPC(id types.Specifier, stream *gateway.Stream, origin *Peer) error {
+func (s *Syncer) handleRPC(ctx context.Context, id types.Specifier, stream *gateway.Stream, origin *Peer) error {
 	switch r := gateway.ObjectForID(id).(type) {
 	case *gateway.RPCShareNodes:
-		peers, err := s.pm.Peers(s.shutdownCtx)
+		peers, err := s.pm.Peers(ctx)
 		if err != nil {
 			return fmt.Errorf("failed to fetch peers: %w", err)
 		} else if n := len(peers); n > 10 {
@@ -260,7 +261,7 @@ func (s *Syncer) handleRPC(id types.Specifier, stream *gateway.Stream, origin *P
 		if _, ok := s.cm.State(bid); ok {
 			return nil // already seen
 		} else if bid.CmpWork(cs.ChildTarget) < 0 {
-			return s.ban(origin, errors.New("peer sent header with insufficient work"))
+			return s.ban(ctx, origin, errors.New("peer sent header with insufficient work"))
 		} else if r.Header.ParentID != s.cm.Tip().ID {
 			// block extends a sidechain, which peer (if honest) believes to be the
 			// heaviest chain
@@ -273,7 +274,7 @@ func (s *Syncer) handleRPC(id types.Specifier, stream *gateway.Stream, origin *P
 			s.log.Warn("couldn't retrieve new block after header relay", zap.Stringer("header", r.Header.ID()), zap.Stringer("origin", origin), zap.Error(err))
 			return nil
 		} else if err := s.cm.AddBlocks([]types.Block{b}); err != nil {
-			return s.ban(origin, err)
+			return s.ban(ctx, origin, err)
 		}
 		s.relayHeader(r.Header, origin) // non-blocking
 		return nil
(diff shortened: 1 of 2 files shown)


lukechampine:
This feels unnecessary to me. Contexts are for operations that might take a while; `PeerStore` operations should never take more than a few milliseconds. Adding a context parameter fundamentally changes the semantics of an interface from "this will block until it completes" to "this may return early, in which case you can either give up or try again." Which means every user of that interface now has to distinguish between those error types and decide what to do. You *also* have to decide what a reasonable timeout would be; 5 seconds sounds quite generous to me, but who knows? Maybe someone will implement a `PeerStore` that uploads a new 30-of-50 sector to Sia for each update, and 5 seconds will be insufficient. :P The point is that we've taken a nice, clean interface and added a bunch of ambiguity to it, not to mention the visual pollution of smearing `context.Context` throughout all the affected code.

Am I correct in assuming that this is motivated by `renterd`'s `PeerStore` implementation being backed by a database, which takes contexts on its methods? If so, I would suggest using `context.Background()` inside the `PeerStore` methods. Even a 5 second timeout would be weird, because (as mentioned above) that would mean the syncer now has to decide whether to retry or abandon each `PeerStore` operation. 

In general, my stances is that contexts should only be used (for cancellation, at least) when there is a) an *observed* performance or UX issue arising from an uncanceled process (e.g. the user hits Ctrl-C and the program doesn't exit within 3 seconds) *and* b) there is no other means of cancellation (e.g. calling `Close` on a `net.Conn`). In this case, I can imagine a `PeerStore` operation taking a while on a database because the database is busy serving some other query. So cancel *that* query, the slow one! Once it's canceled, the `PeerStore` operation will finish quickly, and shutdown (or w/e) can continue.

peterjan:
> This feels unnecessary to me. Contexts are for operations that might take a while; `PeerStore` operations should never take more than a few milliseconds. Adding a context parameter fundamentally changes the semantics of an interface from "this will block until it completes" to "this may return early, in which case you can either give up or try again." Which means every user of that interface now has to distinguish between those error types and decide what to do. You _also_ have to decide what a reasonable timeout would be; 5 seconds sounds quite generous to me, but who knows? Maybe someone will implement a `PeerStore` that uploads a new 30-of-50 sector to Sia for each update, and 5 seconds will be insufficient. :P The point is that we've taken a nice, clean interface and added a bunch of ambiguity to it, not to mention the visual pollution of smearing `context.Context` throughout all the affected code.
> 
> Am I correct in assuming that this is motivated by `renterd`'s `PeerStore` implementation being backed by a database, which takes contexts on its methods? If so, I would suggest using `context.Background()` inside the `PeerStore` methods. Even a 5 second timeout would be weird, because (as mentioned above) that would mean the syncer now has to decide whether to retry or abandon each `PeerStore` operation.
> 
> In general, my stance is that contexts should only be used (for cancellation, at least) when two conditions are satisfied:
> 
> 1. There is an _observed_ performance or UX issue arising from an uncanceled process (e.g. the user hits Ctrl-C and the program doesn't exit within 3 seconds), and
> 2. There is no other means of cancellation (e.g. calling `Close` on a `net.Conn`).
> 
> In this case, I can imagine a `PeerStore` operation taking a while on a database because the database is busy serving some other query. So cancel _that_ query, the slow one! Once it's canceled, the `PeerStore` operation will finish quickly, and shutdown (or w/e) can continue.

I was more or less expecting either a swift merge or pushback on this PR... You are indeed correct this PR was motivated by a change I want to make on `renterd` side where I effectively start respecting the context we pass to our SQL store. You are also right in saying I can easily use `context.Background()` but generally using that is a bit of a code smell for me at least. Definitely appropriate in some cases though, maybe this is one.

I appreciated your comment on how adding a context changes the semantics of an interface, I never really looked at it the way you explained it to me just now. I definitely do think about an interface when I have to design one, but adding a context never really made me think about the ambiguity or extra question you make the caller answer. That's definitely fair and a good point. (You can also see it proving my point by the way, I might be implementing the `PeerStore` so it should probably be my choice whether to bother the eventual caller with those choices yes or no)

I'm still somewhat torn between accepting your point of view and somewhat hardheadedly just keep believing what I always thought was a decent approach, which is to always pass a context to interfaces that might be connected to over a network or a "distance" ... databases obviously being one of those. I might be implementing this peer store on a mongo cluster and be stuck waiting on a 30s connection timeout on shutdown, or anything really, a database lock, a retry mechanism, network failure... essentially anything that made you write "should not take more than a few milliseconds" instead of "do not take more than a few milliseconds"

Like I said though, don't mind closing it and using `context.Background`. The lack of contexts really only bothers me in the cases you would add one.

ChrisSchinnerl (approved):
I don't think this change is very necessary right now in terms of fixing any issues we are seeing in production.
However, we are talking about an interface for a potentially distributed database here, transaction retries happen every now and then and they can block for multiple seconds if we are very unlucky. Cause I don't expect the peer store to get enough interaction to start causing issues in the db. But I don't mind merging this. 

@n8maninger do you want to be the tiebreaker?

n8maninger:
I don't particularly like the pollution `context.Context` requires, but I'm usually in favor of adding `context.Context` anywhere that has the potential to cross a network boundary. The interface consumer should be aware that a call may take some time and be able to decide what is acceptable and whether to retry or fail. Having no way to exit gracefully from a (potentially) indefinite block is also bad interface design.

lukechampine:
*Any* interface can potentially block indefinitely. How do we decide which ones need cancellation?
The bitter truth is that, in API design, there are no universally-applicable rules; when the rubber hits the road, you just have to go with your gut. :^)

In any case, it seems that Nate, Chris, and I all agree on at least one thing: we do not have enough information yet to determine the benefit of a change like this. That is, we have not yet observed any problems in the Syncer arising from a slow `PeerStore`. I suggest that we punt on this until such a problem does arise.

peterjan:
> In the context of this PR, we may need to refine where the Syncer benefits from context cancellation rather than punting it all the way up the stack. For instance, I'm not sure handleRPC benefits from a context param.

I was on the fence about that too, but the following `NOTE` made me feel better about it. (Especially since we have also a `MaxInflightRPCs` that defaults to `3` I figured putting a upper limit on that is not the worst thing.):

```go
// NOTE: we do not set any deadlines on the stream. If a peer is
// slow, fine; we don't need to worry about resource exhaustion
// unless we have tons of peers.
```

> In any case, it seems that Nate, Chris, and I all agree on at least one thing:

I never disagreed on that either, for me this discussion is completely unrelated to _our_ implementation of the `PeerStore` and potential problems that may or may not arise in our implementation. It's an interface design question so its' a question of whether we want to offer the flexibility to others, at the cost of (in our case probably needlessly) polluting the interface.

I don't mind closing this out though. I don't even really need it, aside from the fact that I now won't be able to cleanly pass my contexts all the way down into the store. I understand that merging it without seeing a clear upside is a little weird but to me the interface is cleaner. Regardless of whether it's of use to us.

peterjan:
Ok, I yield

PR #40, opened 2024-03-20
Update `gocritic` enabled/disabled checks to fix linter

peterjan:
This PR adjusts our linter so it passes again after upgrading `golangci` to `v1.57.0`.

### DIFF
--- a/chain/db.go
+++ b/chain/db.go
@@ -402,14 +402,13 @@ func (db *DBStore) deleteFileContractExpiration(id types.FileContractID, windowE
 	key := db.encHeight(windowEnd)
 	val := append([]byte(nil), b.getRaw(key)...)
 	for i := 0; i < len(val); i += 32 {
-		if *(*types.FileContractID)(val[i:]) != id {
-			continue
+		if *(*types.FileContractID)(val[i:]) == id {
+			copy(val[i:], val[len(val)-32:])
+			val = val[:len(val)-32]
+			i -= 32
+			b.putRaw(key, val)
+			return
 		}
-		copy(val[i:], val[len(val)-32:])
-		val = val[:len(val)-32]
-		i -= 32
-		b.putRaw(key, val)
-		return
 	}
 	panic("missing file contract expiration")
 }
--- a/chain/db.go
+++ b/chain/db.go
@@ -402,13 +402,14 @@ func (db *DBStore) deleteFileContractExpiration(id types.FileContractID, windowE
 	key := db.encHeight(windowEnd)
 	val := append([]byte(nil), b.getRaw(key)...)
 	for i := 0; i < len(val); i += 32 {
-		if *(*types.FileContractID)(val[i:]) == id {
-			copy(val[i:], val[len(val)-32:])
-			val = val[:len(val)-32]
-			i -= 32
-			b.putRaw(key, val)
-			return
+		if *(*types.FileContractID)(val[i:]) != id {
+			continue
 		}
+		copy(val[i:], val[len(val)-32:])
+		val = val[:len(val)-32]
+		i -= 32
+		b.putRaw(key, val)
+		return
 	}
 	panic("missing file contract expiration")
 }
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -513,7 +513,8 @@ func (m *Manager) revalidatePool() {
 		delete(m.txpool.indices, txid)
 	}
 	m.txpool.ms = consensus.NewMidState(m.tipState)
-	txns := append(m.txpool.txns, m.txpool.lastReverted...)
+	txns := m.txpool.txns
+	txns = append(txns, m.txpool.lastReverted...)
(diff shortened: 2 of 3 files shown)


chain/manager.go
```diff
@@ -513,7 +513,8 @@ func (m *Manager) revalidatePool() {
 		delete(m.txpool.indices, txid)
 	}
 	m.txpool.ms = consensus.NewMidState(m.tipState)
-	txns := append(m.txpool.txns, m.txpool.lastReverted...)
+	txns := m.txpool.txns
```
ChrisSchinnerl, line 516:
```go
var txns []transactions
txns = append(txns, m.txpool.tns)
txns = append(txns, m.txpool.lastReverted)
```

is probably a bit safer. `append` might reuse the memory of the thing you append to so starting with an empty slice is probably better. Same for potential other spots.

lukechampine: reusing memory is intentional here; this is a "filter" pattern where we want to discard the invalid transactions present in `m.txpool.txns`. We do that by reslicing to 0 and appending.

Usually, the filter pattern looks like this:
```go
rem := m.txpool.txns[:0]
for _, txn := range m.txpool.txns {
    if valid(txn) {
        rem = append(rem, txn)
    } 
}
m.txpool.txns = rem
```
But in this case, since we want to process `lastReverted` as well, I switched things up a little.

If we must keep this silly linter rule, we can rewrite this as:
```go
m.txpool.txns = append(m.txpool.txns, m.txpool.lastReverted...)
rem := m.txpool.txns[:0]
for _, txn := range m.txpool.txns {
    if valid(txn) {
        rem = append(rem, txn)
    } 
}
m.txpool.txns = rem
```
or we could duplicate the loop, i.e. loop over `m.txpool.txns`, then over `m.txpool.lastReverted`. This would be more efficient (since it saves an append), but more verbose.

chain/manager.go
```diff
@@ -806,15 +808,17 @@ func (m *Manager) TransactionsForPartialBlock(missing []types.Hash256) (txns []t
 	for _, txn := range m.txpool.txns {
 		if h := txn.FullHash(); want[h] {
 			txns = append(txns, txn)
-			if delete(want, h); len(want) == 0 {
+			delete(want, h)
+			if len(want) == 0 {
```
lukechampine, line 812: I like this pattern:
```go
if change x; check something about x {
    // ...
}
```
e.g.
```go
if i++; i > 10 {
```
or:
```go
if peers = append(peers, p); len(p) > 10 {
```
I think it follows naturally from the `if err := foo(); err != nil` pattern. It gets a little weird with `delete` because `delete` doesn't return a value, but the principle is the same.

This isn't one of the hills that I care enough to die on, though. :P Just wanted to explain the reasoning here.

chain/db.go
```diff
@@ -402,13 +402,14 @@ func (db *DBStore) deleteFileContractExpiration(id types.FileContractID, windowE
...
-			b.putRaw(key, val)
-			return
+		if *(*types.FileContractID)(val[i:]) != id {
+			continue
 		}
+		copy(val[i:], val[len(val)-32:])
+		val = val[:len(val)-32]
+		i -= 32
+		b.putRaw(key, val)
+		return
```
lukechampine, line 412: I don't see how this is an improvement. I understand the general principle of "put the smaller code block inside the conditional," but when the code blocks in question are <5 lines, it becomes more subjective. In this case, we end up with *more* total lines, plus the meaning of the code changes from "look at an element; if it's the one we want, return it" to "look at an element; if it's not the one we want, skip; otherwise, return it" which feels less clear to me.

lukechampine:
sorry to be a curmudgeon but I disagree with all three of these style rules lmao 😅

peterjan:
> curmudgeon

Haha 'curmudgeon', I don't mind at all Luke. I only wanted to apply the same linters we have in `renterd` because Chris told me we'll move towards centralising these configs soon. Since the changes were minimal I figured to just open the PR. For what it's worth though, I kind of like `initClause`. The other two I don't really care about all that much, especially `nestingReduce`.

ChrisSchinnerl:
To be fair the fact that it wasn't clear whether memory should be reused or not from a quick glance is probably enough reason to keep that linter rule and use the more traditional pattern to make that clear.

peterjan:
I don't want to make this PR the PR where we discuss what checks we should enable/disable. Can we do that when we centralise the configs? I reverted all changes only to make this PR all about getting `master` to pass CI.

ChrisSchinnerl (changes requested):
I think we should still keep the append linter but I don't mind dropping the other ones. 

Definitely not in favour of running linters only once a month cause
1. nobody ever checks the result of scheduled actions
2. linting once a month is like saying you'll clean your apartment once a month instead of putting stuff away right away. You'll just build up a mess that is too annoying to fix.

ChrisSchinnerl (approved):

PR #45, opened 2024-04-16
Cache ephemeral utxo values

peterjan:
Not 100% sure here but I think it's ok to do this. If we ignore all ephemeral elements and thus don't cache their utxo value, I run into the panic on line 194 where we're calculating the total inflow.

### DIFF
--- a/wallet/update.go
+++ b/wallet/update.go
@@ -158,12 +158,18 @@ func applyChainState(tx UpdateTx, address types.Address, cau chain.ApplyUpdate) 
 	utxoValues := make(map[types.SiacoinOutputID]types.Currency)
 
 	cau.ForEachSiacoinElement(func(se types.SiacoinElement, spent bool) {
-		if se.SiacoinOutput.Address != address || ephemeral[types.Hash256(se.ID)] {
+		if se.SiacoinOutput.Address != address {
 			return
 		}
 
 		// cache the value of the utxo to use when calculating outflow
 		utxoValues[types.SiacoinOutputID(se.ID)] = se.SiacoinOutput.Value
+
+		// ignore ephemeral elements
+		if ephemeral[types.Hash256(se.ID)] {
+			return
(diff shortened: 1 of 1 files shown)


wallet/update.go
```diff
@@ -158,12 +158,18 @@ func applyChainState(tx UpdateTx, address types.Address, cau chain.ApplyUpdate)
 	utxoValues := make(map[types.SiacoinOutputID]types.Currency)
 
 	cau.ForEachSiacoinElement(func(se types.SiacoinElement, spent bool) {
-		if se.SiacoinOutput.Address != address || ephemeral[types.Hash256(se.ID)] {
+		if se.SiacoinOutput.Address != address {
 			return
 		}
 
 		// cache the value of the utxo to use when calculating outflow
 		utxoValues[types.SiacoinOutputID(se.ID)] = se.SiacoinOutput.Value
```
n8maninger, line 166: It'd be better to move this up above the ephemeral check instead. We could pass the ephemeral outputs to the data store, but then we would need to pass that context to the store so it can handle them properly. If you're doing a balance calculation, that can get annoying.

wallet/update.go
```diff
@@ -158,12 +158,18 @@ func applyChainState(tx UpdateTx, address types.Address, cau chain.ApplyUpdate)
...
+
+		// ignore ephemeral elements
+		if ephemeral[types.Hash256(se.ID)] {
```
lukechampine, line 169: unnecessary cast (`se.ID` is already a `types.Hash256`)

lukechampine:
looks fine to me (irrespective of Nate's proposed change)

lukechampine (approved):

wallet/update.go
n8maninger: Disregard this. Missed that the check still exists

n8maninger (approved):

PR #46, opened 2024-04-17
fix invalid MinerPayouts in testutil.MineBlock

peterjan:
Silly little bug in our `MineBlock` implementation where we call `PoolTransactions` twice and thus introduce a race. This eventually results in an error upon validating the block that indicates the miner payout sum doesn't match block reward + fees.

### DIFF
--- a/miner.go
+++ b/miner.go
@@ -38,7 +38,14 @@ func FindBlockNonce(cs consensus.State, b *types.Block, timeout time.Duration) b
 // MineBlock constructs a block from the provided address and the transactions
 // in the txpool, and attempts to find a nonce for it that meets the PoW target.
 func MineBlock(cm *chain.Manager, addr types.Address, timeout time.Duration) (types.Block, bool) {
+retry:
 	cs := cm.TipState()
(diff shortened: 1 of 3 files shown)


lukechampine (changes requested):
we should support V2 transactions while we're at it.

Also, I spot another race: the state when you call `cm.PoolTransactions` might not match the state when you called `cm.TipState`! That's annoying. We can squash it with a loop:
```go
retry:
    cs := cm.TipState()
    txns := cm.PoolTransactions()
    txnsv2 := cm.PoolTransactionsV2()
    if cs.Index != cm.Tip() {
        goto retry
    }
```
But it might make more sense to add a helper function to `chain.Manager` that returns a state and the pool transactions together.

Alternatively, maybe this isn't a big deal? Miners already have to assume that a new block might arrive while they're mining. So they need to regularly check the tip and restart anyway.

lukechampine:
Indeed. Though, that version also needs the race fix...

lukechampine (approved):

PR #54, opened 2024-05-26
Add support for v2 transactions to Single Address Wallet

### LINKED ISSUE #49: Add support for V2 types to SingleAddressWallet

n8maninger:
Adds support for v2 events and constructing v2 transactions to the single address wallet.

Closes #49

### DIFF
--- a/testutil/wallet.go
+++ b/testutil/wallet.go
@@ -44,14 +44,6 @@ func (et *ephemeralWalletUpdateTxn) UpdateStateElements(elements []types.StateEl
 	return nil
 }
 
-func (et *ephemeralWalletUpdateTxn) AddEvents(events []wallet.Event) error {
-	et.store.events = append(events, et.store.events...)
-	sort.Slice(et.store.events, func(i, j int) bool {
-		return et.store.events[i].MaturityHeight < et.store.events[j].MaturityHeight
-	})
-	return nil
-}
-
 func (et *ephemeralWalletUpdateTxn) ApplyIndex(index types.ChainIndex, created, spent []types.SiacoinElement, events []wallet.Event) error {
 	for _, se := range spent {
 		if _, ok := et.store.utxos[types.SiacoinOutputID(se.ID)]; !ok {
--- a/testutil/wallet.go
+++ b/testutil/wallet.go
@@ -46,6 +46,9 @@ func (et *ephemeralWalletUpdateTxn) UpdateStateElements(elements []types.StateEl
 
 func (et *ephemeralWalletUpdateTxn) AddEvents(events []wallet.Event) error {
 	et.store.events = append(events, et.store.events...)
+	sort.Slice(et.store.events, func(i, j int) bool {
+		return et.store.events[i].MaturityHeight < et.store.events[j].MaturityHeight
+	})
 	return nil
 }
 
@@ -137,6 +140,7 @@ func (es *EphemeralWalletStore) UnspentSiacoinElements() (utxos []types.SiacoinE
 	defer es.mu.Unlock()
 
 	for _, se := range es.utxos {
+		se.MerkleProof = append([]types.Hash256(nil), se.MerkleProof...)
 		utxos = append(utxos, se)
 	}
 	return utxos, nil
--- a/wallet/events.go
+++ b/wallet/events.go
@@ -48,6 +48,17 @@ type (
 		Missed         bool                               `json:"missed"`
 	}
 
+	// EventV1Transaction is a transaction event that includes the transaction
+	EventV1Transaction types.Transaction
+
+	// EventV2Transaction is a transaction event that includes the transaction
+	EventV2Transaction types.V2Transaction
+
+	// EventData contains the data associated with an event.
+	EventData interface {
+		isEvent() bool
+	}
+
 	// An Event is a transaction or other event that affects the wallet including
 	// miner payouts, siafund claims, and file contract payouts.
 	Event struct {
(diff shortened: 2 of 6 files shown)


peterjan (dismissed):
Solid 👍 one nit I considered was that I preferred "source" over "type" but other than that it LGTM

wallet/events.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// An Event is a transaction or other event that affects the wallet including
+	// miner payouts, siafund claims, and file contract payouts.
+	Event struct {
+		ID             types.Hash256    `json:"id"`
+		Index          types.ChainIndex `json:"index"`
+		Inflow         types.Currency   `json:"inflow"`
+		Outflow        types.Currency   `json:"outflow"`
+		Type           string           `json:"type"`
+		Data           any              `json:"data"`
```
peterjan, line 59: `nit` Feel free to ignore but I think we would generally use `interface{}` here, same thing of course.

wallet/update.go
```diff
@@ -36,98 +45,207 @@ type (
...
+
+		switch data := data.(type) {
+		case EventMinerPayout:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeMinerPayout
+			ev.MaturityHeight = cs.MaturityHeight()
+		case EventFoundationSubsidy:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeFoundationSubsidy
+			ev.MaturityHeight = cs.MaturityHeight()
```
peterjan, line 79: `nit` might as well make this a convenience variable same as we do with the index

wallet/update.go
```diff
@@ -36,98 +45,207 @@ type (
...
+		case EventV2ContractPayout:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeV2Contract
+			ev.MaturityHeight = cs.MaturityHeight()
+		}
+
+		events = append(events, ev)
+	}
+
+	relevantV1Txn := func(txn types.Transaction) bool {
```
peterjan, line 121: wondering whether it's worth adding a `Contains` on some of these types for which we keep writing `isRelevantX` for

wallet/wallet.go
```diff
@@ -373,6 +339,21 @@ func (sw *SingleAddressWallet) FundTransaction(txn *types.Transaction, amount ty
...
+
+// FundTransaction adds siacoin inputs worth at least amount to the provided
+// transaction. If necessary, a change output will also be added. The inputs
+// will not be available to future calls to FundTransaction unless ReleaseInputs
+// is called.
+func (sw *SingleAddressWallet) FundTransaction(txn *types.Transaction, amount types.Currency, useUnconfirmed bool) ([]types.Hash256, error) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	selected, inputSum, err := sw.selectUTXOs(amount, len(txn.SiacoinInputs), useUnconfirmed)
```
peterjan, line 353: might want to return early if amount is zero

wallet/update.go
```diff
@@ -36,98 +45,207 @@ type (
...
+		case EventMinerPayout:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeMinerPayout
+			ev.MaturityHeight = cs.MaturityHeight()
+		case EventFoundationSubsidy:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeFoundationSubsidy
+			ev.MaturityHeight = cs.MaturityHeight()
+		case types.Transaction:
+			for _, si := range data.SiacoinInputs {
```
ChrisSchinnerl, line 81: I think you can call `relevantV1Txn` here if you move it up.

wallet/update.go
```diff
@@ -36,98 +45,207 @@ type (
...
+				}
+			}
+			ev.MaturityHeight = index.Height
+			ev.Type = EventTypeV1Transaction
+		case EventV1ContractPayout:
+			ev.Inflow = data.SiacoinElement.SiacoinOutput.Value
+			ev.Type = EventTypeV1Contract
+			ev.MaturityHeight = cs.MaturityHeight()
+		case types.V2Transaction:
+			for _, si := range data.SiacoinInputs {
```
ChrisSchinnerl, line 99: Same but with `relevantV2Txn`

ChrisSchinnerl (approved):

wallet/events.go
ChrisSchinnerl: I think we have only used `interface{}` in most cases because we supported pre-"any" go versions. 

Something we could do for hardening though is have a `EventData` interface with some method `IsEventData` that doesn't return anything. Just to get some extra type safety.

testutil/wallet.go
```diff
@@ -46,6 +46,9 @@ func (et *ephemeralWalletUpdateTxn) UpdateStateElements(elements []types.StateEl
 
 func (et *ephemeralWalletUpdateTxn) AddEvents(events []wallet.Event) error {
 	et.store.events = append(events, et.store.events...)
+	sort.Slice(et.store.events, func(i, j int) bool {
```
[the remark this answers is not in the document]
ChrisSchinnerl: +1

wallet/update.go
ChrisSchinnerl: I had the same thought. Obviously it would need to be in `core` but we essentially have the same code in every piece of node software.

wallet/update.go
n8maninger: I'd prefer not to loop twice if I can help it. I might refactor relevantV1Txn..

wallet/update.go
n8maninger: ditto

wallet/update.go
n8maninger: I don't think it's necessary. The logic is simple and there may be optimizations for exiting early that would be ignored if we added a helper.

testutil/wallet.go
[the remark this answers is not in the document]
n8maninger: I removed the sort in favor of the existing stable sort in `WalletEvents(offset, limit int)`. Marking as resolved unless we want to talk more about transaction ordering.

ChrisSchinnerl (approved):

peterjan (approved):

lukechampine (approved):

PR #56, opened 2024-05-31
Remove `Data` from `Event`

ChrisSchinnerl:
Considering the `SingleAddressWallet` is not meant for an explorer and only tracks events to display the inflow and outflow of siacoins, would it be fine to drop the `Data` from the `Event` @n8maninger ? Seems like the only purpose it has is to initialise the `Event` type right now.


n8maninger:
👎

PR #70, opened 2024-07-09
Use context for OnReorg

n8maninger:
Updates `OnReorg` to use `context.Context` for managing the lifecycle of the registered functions. Now, the function will be removed when the context is canceled instead of needing to keep track of an extra cancellation function. 

Builds on https://github.com/SiaFoundation/core/pull/180

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -1,6 +1,7 @@
 package chain
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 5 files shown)


lukechampine:
Looks like `walletd` is currently the only consumer of `OnReorg`, in `NewManager`. So `NewManager` would have to take a `context.Context`? Feels weird.

n8maninger:
This builds on something I was playing around with over the weekend. Passing the context in was less annoying than tracking an additional unsubscribe func. Personally feels more idiomatic too, but I don’t have a strong opinion one way or the other.

For `walletd`, there’s an existing context in the sync loop goroutine that could be used rather than passing one in.

ChrisSchinnerl (approved):

PR #73, opened 2024-07-25
Refactor SingleAddressWallet store

n8maninger:
Removes `UpdateChainState` as a method on the Store interface to make the SingleAddressWallet slightly less opinionated about how its receiving chain updates from the backing store.


peterjan (approved):
This cleans things up in `renterd` as well. PR LGTM 👍

PR #76, opened 2024-07-30
Always include v2 block data after allow height

n8maninger:
Updating the transaction pool basis fails after the allow height if any block in the chain does not contain v2 data. This is a quick fix for unit tests. We can discuss whether the tpool behavior is fine in a follow up.

https://github.com/SiaFoundation/coreutils/blob/872e1d92e9fce751a31196b0921b44f76300710a/chain/manager.go#L1059

### DIFF
--- a/miner.go
+++ b/miner.go
@@ -54,6 +54,13 @@ retry:
 			Address: addr,
 		}},
 	}
+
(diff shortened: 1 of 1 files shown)


lukechampine (approved):
ah, I see why it fails. I think the current txpool behavior is just wrong. I'll make a PR to fix it.

PR #77, opened 2024-07-31
syncer: Add Close and ensure goroutines exit

lukechampine:
See https://github.com/SiaFoundation/renterd/pull/1407

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -276,6 +276,8 @@ func (s *Syncer) runPeer(p *Peer) error {
 	}()
 
 	inflight := make(chan struct{}, s.config.MaxInflightRPCs)
+	var wg sync.WaitGroup
+	defer wg.Wait()
 	for {
 		if p.Err() != nil {
 			return fmt.Errorf("peer error: %w", p.Err())
@@ -286,7 +288,9 @@ func (s *Syncer) runPeer(p *Peer) error {
 			return fmt.Errorf("failed to accept rpc: %w", err)
 		}
 		inflight <- struct{}{}
+		wg.Add(1)
 		go func() {
+			defer wg.Done()
 			defer stream.Close()
 			// NOTE: we do not set any deadlines on the stream. If a peer is
 			// slow, fine; we don't need to worry about resource exhaustion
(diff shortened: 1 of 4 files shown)


syncer/syncer.go
```diff
@@ -276,6 +276,8 @@ func (s *Syncer) runPeer(p *Peer) error {
 	}()
 
 	inflight := make(chan struct{}, s.config.MaxInflightRPCs)
+	var wg sync.WaitGroup
+	defer wg.Done()
```
peterjan, line 280: This should be `defer wg.Wait()`

peterjan (commented):
I didn't immediately see how this PR was ensuring the `syncLoop` exits, would it clean things up if we added a `wg` to the `syncer` and then use that in all loops and replace the `<-errChan` with `wg.Wait` right in `Run`?

syncer/syncer.go
lukechampine: whoops, you're right

lukechampine:
I think that would make the error handling more annoying; instead of sending the errors down a channel, you would need separate variables or a slice or something.

peterjan (dismissed):

peterjan:
> I think that would make the error handling more annoying; instead of sending the errors down a channel, you would need separate variables or a slice or something.

I'd keep the err channel but remove the two consecutive blocks where we don't care about the `err` being returned anyway. Anyway, doesn't matter, LGTM 👍

syncer/syncer.go
```diff
@@ -632,7 +628,7 @@ func (s *Syncer) syncLoop() error {
 // Run spawns goroutines for accepting inbound connections, forming outbound
 // connections, and syncing the blockchain from active peers. It blocks until an
 // error occurs, upon which all connections are closed and goroutines are
-// terminated. To gracefully shutdown a Syncer, close its net.Listener.
+// terminated.
 func (s *Syncer) Run() error {
```
n8maninger, line 632: I think you can get rid of the `shutdownCtx` and `shutdownCtxCancel` fields by moving it here. Has, essentially, the same effect now that the waitgroup is there and is more idiomatic.

```go
func (s *Syncer) Run(ctx context.Context) error {
	errChan := make(chan error)
	go func() { errChan <- s.acceptLoop(ctx) }()
	go func() { errChan <- s.peerLoop(ctx) }()
	go func() { errChan <- s.syncLoop(ctx) }()
	err := <-errChan
        // ....
}
```

n8maninger (dismissed):

n8maninger (approved):

PR #83, opened 2024-08-08
Add anagami testnet parameters

ChrisSchinnerl:
@n8maninger can we also add the bootstrap peers? Seems like unnecessary room for inconsistency not to have them in a central location as well.


n8maninger:
Bootstrap peers are in the syncer package

n8maninger (approved):

PR #89, opened 2024-08-23
Fix tpool race

n8maninger:
Fixes a race condition primarily seen when exchanging v2 transaction sets in RHP4. The renter sends the host a transaction set with a particular basis. If the tip changes before the RPC completes and the transaction includes ephemeral outputs, the transaction set will be invalidated and the RPC will fail. This fixes that race by changing the tpool to update the ephemeral state elements and proofs when updating the transaction set basis.

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -928,13 +928,23 @@ func (m *Manager) UnconfirmedParents(txn types.Transaction) []types.Transaction 
 	return parents
 }
 
-// V2UnconfirmedParents returns the v2 transactions in the txpool that are referenced
-// by txn.
-func (m *Manager) V2UnconfirmedParents(txn types.V2Transaction) []types.V2Transaction {
+// V2TransactionSet returns the full transaction set and basis necessary for
+// broadcasting a transaction. If the provided basis does not match the current
+// tip the transaction will be updated. The transaction set includes the parents
+// and the transaction itself in an order valid for broadcasting.
+func (m *Manager) V2TransactionSet(basis types.ChainIndex, txn types.V2Transaction) (types.ChainIndex, []types.V2Transaction, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.revalidatePool()
 
+	// update the transaction's basis to match tip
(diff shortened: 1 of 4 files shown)


wallet/update.go
```diff
@@ -371,15 +371,25 @@ func (sw *SingleAddressWallet) UpdateChainState(tx UpdateTx, reverted []chain.Re
 			ID:     cru.Block.ID(),
 			Height: cru.State.Index.Height + 1,
 		}
-		if err := revertChainUpdate(tx, revertedIndex, sw.addr, cru); err != nil {
-			return err
+		sw.mu.Lock()
+		err := revertChainUpdate(tx, revertedIndex, sw.addr, cru)
+		if err != nil {
```
ChrisSchinnerl, line 376: At this point we could make `revertChainUpdate` and `applyChainState` methods `SingleAddressWallet` since they need to lock it and update its field.

ChrisSchinnerl (dismissed):

wallet/update.go
n8maninger: That’s true. Do you see a reason to keep UpdateChainState in that instance?

chain/manager.go
```diff
@@ -928,13 +928,23 @@ func (m *Manager) UnconfirmedParents(txn types.Transaction) []types.Transaction
...
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.revalidatePool()
 
+	// update the transaction's basis to match tip
+	_, txns, err := m.updateV2TransactionSet(basis, []types.V2Transaction{txn})
+	if err != nil {
+		return types.ChainIndex{}, nil, fmt.Errorf("failed to update transaction set basis: %w", err)
+	}
+	txn = txns[0]
```
peterjan, line 945: Shouldn't we handle the case where `txns` comes back empty? If not I would still go for a sanity check and error out here.

wallet/update.go
ChrisSchinnerl: I think it's fine to keep it since `UpdatesSince` also returns both apply and revert updates. Saves us the loops in `renterd` and `hostd`. But I'm not too attached.

ChrisSchinnerl (approved):

PR #93, opened 2024-09-01
Always allow manual peer connections

n8maninger:
Removes the allow check from the public `Connect` method on the syncer. This will enable implementers to manually connect to peers regardless of the current number of inbound or outbound connections.


ChrisSchinnerl (approved):

peterjan (approved):
LGTM 👍

PR #96, opened 2024-09-13
Implement rhp4 basic client and server

n8maninger:
Requires https://github.com/SiaFoundation/core/pull/200

Need to readd the robust RPC client on top of the base implementations in a f/u


rhp/v4/server.go
```diff
@@ -0,0 +1,1139 @@
...
+		return func() {
+			s.log.Debug("RPCWriteSector", zap.String("context", context), zap.Duration("duration", time.Since(start)))
+		}
+	}
+	log := elapsed("write sector streaming")
+	var req rhp4.RPCWriteSectorStreamingRequest
+	if err := rhp4.ReadRequest(stream, &req); err != nil {
+		return errorDecodingError("failed to read request: %v", err)
+	}
+	log()
```
ChrisSchinnerl, line 219: reminder to tone down the logging a bit before merging

rhp/v4/rpc_test.go
```diff
@@ -0,0 +1,1510 @@
...
+	revised, _, err := rhp4.RPCFundAccounts(transport, cs, renterKey, revision, []proto4.AccountDeposit{
+		{Account: account, Amount: accountFundAmount},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	revision.Revision = revised
+
+	// renew the contract
+	renewal := proto4.ReviseForRenewal(revision.Revision, settings.Prices, revision.Revision.ProofHeight+10, revision.Revision.ExpirationHeight+10, types.Siacoins(50), types.Siacoins(100))
```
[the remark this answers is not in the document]
n8maninger: I might be misunderstanding the scenario, but the UTXOs from a renewed contract are created immediately with a maturity height. 

If I’m understanding correctly, the host funds from the initial contract will be created and the host won’t lock any collateral into the new contract.

chain/hostannouncement.go
```diff
@@ -12,29 +13,57 @@ const attestationHostAnnouncement = "HostAnnouncement"
...
+	}
+
+	// A NetAddress is a pair of protocol and address that a host may be reached on
+	NetAddress struct {
+		Protocol string `json:"protocol"`
+		Address  string `json:"address"`
+	}
+
+	// A V2HostAnnouncement lists all the network addresses a host may be reached on
+	V2HostAnnouncement []NetAddress
```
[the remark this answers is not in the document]
n8maninger: Having to scan the settings over tcp to get the web transport protocol feels unnecessary. Renters will be able to see all the different ways to connect to a host beforehand and pick and choose which they care about. It does add more complexity on both ends, but I think it’s worth it.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,529 @@
...
+
+	var resp rhp4.RPCLatestRevisionResponse
+	if err := rhp4.ReadResponse(s, &resp); err != nil {
+		return types.V2FileContract{}, fmt.Errorf("failed to read response: %w", err)
+	}
+	return resp.Contract, nil
+}
+
+// RPCSectorRoots returns the sector roots for a contract
+func RPCSectorRoots(t TransportClient, cs consensus.State, prices rhp4.HostPrices, sk types.PrivateKey, contract ContractRevision, offset, length uint64) (types.V2FileContract, []types.Hash256, error) {
```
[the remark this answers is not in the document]
n8maninger: The only reason I can think of is proof of ownership. I don’t have a problem changing it to an account

rhp/v4/rpc.go
```diff
@@ -0,0 +1,529 @@
...
+	revision.RenterSignature = sk.SignHash(sigHash)
+
+	req := rhp4.RPCFundAccountsRequest{
+		ContractID:      contract.ID,
+		Deposits:        deposits,
+		RenterSignature: revision.RenterSignature,
+	}
+
+	s := t.DialStream()
+	defer s.Close()
```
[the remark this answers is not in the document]
n8maninger: I would prefer to add context to it for cancellation. We may be able to get away with adding it to the transport. This would probably be fine for the mux with no changes:
```go
func (mt *muxTransport) DialStream(ctx context.Context) net.Conn {
  s := mt.mux.DialStream()
  go func() {
	<-ctx.Done()
	s.Close()
  }()
  return s
}
```

Would need to check that closing the stream does actually interupt any calls to read and write.

chain/hostannouncement.go
```diff
@@ -12,29 +13,57 @@ const attestationHostAnnouncement = "HostAnnouncement"
...
+	na.Address = d.ReadString()
 }
 
 // ToAttestation encodes a host announcement as an attestation.
-func (ha HostAnnouncement) ToAttestation(cs consensus.State, sk types.PrivateKey) types.Attestation {
+func (ha V2HostAnnouncement) ToAttestation(cs consensus.State, sk types.PrivateKey) types.Attestation {
+	buf := bytes.NewBuffer(nil)
+	e := types.NewEncoder(buf)
+	types.EncodeSlice(e, ha)
+	e.Flush()
```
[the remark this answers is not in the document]
n8maninger: The Encoder implementation ensures it won't, but we can panic here too I guess. The behavior hasn't really changed in this PR from ca618b0

chain/hostannouncement.go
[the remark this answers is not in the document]
n8maninger: Another thing to consider, this enables IPv4 and IPv6 support on hosts without using a domain.

chain/hostannouncement.go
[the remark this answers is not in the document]
lukechampine: writes to a bytes.Buffer never fail

rhp/v4/rpc.go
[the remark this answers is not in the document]
lukechampine: Mux v2 has `DialStreamContext`

rhp/v4/rpc.go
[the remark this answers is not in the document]
ChrisSchinnerl: Ahhh right I think that might have been it. That makes the use case a bit harder to do but not impossible if we ever need it by just forwarding the proof and signed revision.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,561 @@
...
+
+	// A ChainReader reports the chain state and manages the mempool.
+	ChainReader interface {
+		Tip() types.ChainIndex
+		TipState() consensus.State
+
+		// V2TransactionSet returns the full transaction set and basis necessary for
+		// broadcasting a transaction. If the provided basis does not match the current
+		// tip, the transaction will be updated. The transaction set includes the parents
+		// and the transaction itself in an order valid for broadcasting.
```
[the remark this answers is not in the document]
n8maninger: I believe the existing statement is grammatically correct

rhp/v4/rpc_test.go
```diff
@@ -0,0 +1,1367 @@
...
+	return fs.pk.SignHash(h)
+}
+func (fs *fundAndSign) PublicKey() types.PublicKey {
+	return fs.pk.PublicKey()
+}
+func (fs *fundAndSign) Address() types.Address {
+	return fs.w.Address()
+}
+
+func testRenterHostPair(tb testing.TB, hostKey types.PrivateKey, cm rhp4.ServerChainManager, s rhp4.Syncer, w rhp4.Wallet, c rhp4.Contractor, sr rhp4.SettingsReporter, ss rhp4.SectorStore, log *zap.Logger) rhp4.TransportClient {
```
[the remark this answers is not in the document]
n8maninger: There's a benchmark at the bottom

chain/hostannouncement.go
```diff
@@ -44,9 +75,14 @@ func (ha HostAnnouncement) ToArbitraryData(sk types.PrivateKey) []byte {
 	specifierHostAnnouncement.EncodeTo(e)
 	e.WriteString(ha.NetAddress)
 	sk.PublicKey().UnlockKey().EncodeTo(e)
-	e.Flush()
+	if err := e.Flush(); err != nil {
```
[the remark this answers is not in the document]
n8maninger: To hash the encoded bytes for the signature below

rhp/v4/server.go
```diff
@@ -0,0 +1,1037 @@
...
+
+	switch {
+	case req.DataLength > rhp4.SectorSize:
+		return errorBadRequest("sector size %v exceeds maximum %v", req.DataLength, rhp4.SectorSize)
+	case req.DataLength%rhp4.LeafSize != 0:
+		return errorBadRequest("sector length %v must be a multiple of segment size %v", req.DataLength, rhp4.LeafSize)
+	case req.Duration > settings.MaxSectorDuration:
+		return errorBadRequest("sector duration %v exceeds maximum %v", req.Duration, settings.MaxSectorDuration)
+	case settings.RemainingStorage < rhp4.SectorSize:
+		return rhp4.ErrNotEnoughStorage
```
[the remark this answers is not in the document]
n8maninger: Everything is padded out to a sector by the host.

rhp/v4/server.go
```diff
@@ -0,0 +1,1037 @@
...
+	}
+
+	settings := s.settings.RHP4Settings()
+
+	switch {
+	case req.DataLength > rhp4.SectorSize:
+		return errorBadRequest("sector size %v exceeds maximum %v", req.DataLength, rhp4.SectorSize)
+	case req.DataLength%rhp4.LeafSize != 0:
+		return errorBadRequest("sector length %v must be a multiple of segment size %v", req.DataLength, rhp4.LeafSize)
+	case req.Duration > settings.MaxSectorDuration:
```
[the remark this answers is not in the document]
n8maninger: I could see the benefit of MinSectorDuration. You could get a lot of writes pretty cheaply for storing something for a single block.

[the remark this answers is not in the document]
n8maninger: Greater than zero is probably enough

rhp/v4/rpc.go
```diff
@@ -0,0 +1,561 @@
...
+		RenterInputs:  renterSiacoinElements,
+		RenterParents: formationSet,
+	}
+	if err := rhp4.WriteRequest(s, rhp4.RPCFormContractID, &req); err != nil {
+		return ContractRevision{}, TransactionSet{}, fmt.Errorf("failed to write request: %w", err)
+	}
+
+	var hostInputsResp rhp4.RPCFormContractResponse
+	if err := rhp4.ReadResponse(s, &hostInputsResp); err != nil {
+		return ContractRevision{}, TransactionSet{}, fmt.Errorf("failed to read host inputs response: %w", err)
```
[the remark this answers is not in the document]
n8maninger: Most of the other RPC are single roundtrips. The additional context may be useful for debugging since there are multiple responses to write and read.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,561 @@
...
+
+	// add the host inputs to the transaction
+	var hostInputSum types.Currency
+	for _, si := range hostInputsResp.HostInputs {
+		hostInputSum = hostInputSum.Add(si.Parent.SiacoinOutput.Value)
+		renewalTxn.SiacoinInputs = append(renewalTxn.SiacoinInputs, si)
+	}
+
+	// verify the host added enough inputs
+	if !hostInputSum.Equals(hostCost) {
```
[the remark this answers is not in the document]
n8maninger: That's handled by the setup transaction. It's a consensus rule that transactions with contract renewals can't have SiacoinOutputs for some reason.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,561 @@
...
+		return ContractRevision{}, TransactionSet{}, fmt.Errorf("expected host to fund %v, got %v", hostCost, hostInputSum)
+	}
+
+	// sign the renter inputs
+	signer.SignV2Inputs(&renewalTxn, []int{0})
+	// sign the renewal
```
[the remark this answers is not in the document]
n8maninger: ???

rhp/v4/server.go
```diff
@@ -0,0 +1,1044 @@
...
+	existing := state.Revision
+	if !req.ValidChallengeSignature(existing) {
+		return errorBadRequest("invalid challenge signature")
+	}
+
+	renewal := rhp4.NewRenewal(existing, prices, req.Renewal)
+
+	// create the renewal transaction
+	renewalTxn := types.V2Transaction{
+		MinerFee: req.MinerFee,
```
ChrisSchinnerl, line 656: I noticed that when we validate `req.MinerFee`, we only check that the fee is > 0. In RHP3 we have fields on the price table to make sure we agree on a minimum.

My thought is whether we can't just eliminate that discrepancy altogether now that we no longer sign the full txn and only the contract.

That means, the renter funds the `ContractPrice` but doesn't send a miner fee. Instead, the host just pays for the miner fee. If the renter thinks the amount the host paid isn't enough, it can then just add another input and increase the fees right on its end right? 

cc @lukechampine

n8maninger: Isn't there still a scenario where the host rejects the formation because the renter didn't pay a high enough contract price to broadcast the transaction.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,529 @@
...
+		return types.Hash256{}, fmt.Errorf("failed to write request: %w", err)
+	}
+
+	var resp rhp4.RPCWriteSectorResponse
+	if err := rhp4.ReadResponse(s, &resp); err != nil {
+		return types.Hash256{}, fmt.Errorf("failed to read response: %w", err)
+	}
+
+	// TODO: stream?
+	var sector [rhp4.SectorSize]byte
```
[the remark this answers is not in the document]
n8maninger: Since we have to put in a data len anyway, I think I'm still in agreement with your first comment that its not worth it for uploads.

rhp/v4/server.go
n8maninger: Maybe we can expose the not enough fees error in the tpool, have the host check for that error and accept the contract, the rebroadcast code automatically handles it later when (hopefully) fees are lower.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,529 @@
...
+	copy(sector[:], data)
+	root := rhp4.SectorRoot(&sector)
+	if root != resp.Root {
+		return types.Hash256{}, fmt.Errorf("invalid root returned: expected %v, got %v", root, resp.Root)
+	}
+	return root, nil
+}
+
+// RPCModifySectors modifies sectors on the host
+func RPCModifySectors(t TransportClient, cs consensus.State, prices rhp4.HostPrices, sk types.PrivateKey, contract ContractRevision, actions []rhp4.WriteAction) (types.V2FileContract, error) {
```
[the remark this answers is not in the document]
n8maninger: Should we consider adjusting the RPC on the host side to return the amount spent? The RPC are going to be calling `prices.RPC*Price`, which seems like it could be done outside of the RPC function easily enough and we can ding the host if it doesn't match up.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,718 @@
...
+	hostFormationTxn := hostTransactionSetResp.TransactionSet[len(hostTransactionSetResp.TransactionSet)-1]
+	if len(hostFormationTxn.FileContracts) != 1 {
+		return RPCFormContractResult{}, fmt.Errorf("expected exactly one contract")
+	}
+
+	// check for no funny business
+	formationTxnID := formationTxn.ID()
+	hostFormationTxnID := hostFormationTxn.ID()
+	if formationTxnID != hostFormationTxnID {
+		return RPCFormContractResult{}, fmt.Errorf("expected transaction IDs to match %v != %v", formationTxnID, hostFormationTxnID)
```
lukechampine, line 554: in my experience there's little benefit to printing mismatched hashes; unlike other comparisons, it rarely tells you anything beyond "they are different"

rhp/v4/rpc.go
```diff
@@ -0,0 +1,718 @@
...
+		},
+		Cost: renterCost,
+	}, nil
+}
+
+// RPCRenewContract renews a contract with a host
+func RPCRenewContract(ctx context.Context, t TransportClient, cr ChainReader, signer FormContractSigner, p rhp4.HostPrices, existing types.V2FileContract, params rhp4.RPCRenewContractParams) (RPCRenewContractResult, error) {
+	renewal := rhp4.NewRenewal(existing, p, params)
+	renewalTxn := types.V2Transaction{
+		MinerFee: types.Siacoins(1),
```
lukechampine, line 580: :eyes:

testutil/host.go
```diff
@@ -0,0 +1,395 @@
...
+func (es *EphemeralSectorStore) WriteSector(root types.Hash256, sector *[proto4.SectorSize]byte, expiration uint64) error {
+	es.mu.Lock()
+	defer es.mu.Unlock()
+	es.sectors[root] = *sector
+	return nil
+}
+
+// An EphemeralContractor is an in-memory minimal rhp4.Contractor for testing.
+type EphemeralContractor struct {
+	mu sync.Mutex // protects the fields below
```
lukechampine, line 47: a `mu` that covers all fields conventionally lives at the bottom

testutil/host.go
```diff
@@ -0,0 +1,395 @@
...
+	"crypto/ed25519"
+	"errors"
+	"net"
+	"sync"
+	"testing"
+
+	proto4 "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	rhp4 "go.sia.tech/coreutils/rhp/v4"
```
lukechampine, line 13: I've been writing this as `rhpv2`/`rhpv3` elsewhere; not a big deal, but we should be consistent one way or the other.

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+		RiskedCollateral types.Currency `json:"riskedCollateral"`
+	}
+)
+
+type (
+	// A TransportListener is a generic multiplexer for incoming streams.
+	TransportListener interface {
+		AcceptStream() (net.Conn, error)
+		Close() error
+	}
```
lukechampine, line 43: this makes me wonder if `mux` should implement `net.Listener` 🤔. But I suppose then you have to return a `net.Conn` instead of a `mux.Stream`, which kinda sucks.

lukechampine: Anyway I would call this `Mux`

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+	case rhp4.RPCReadSectorID:
+		err = s.handleRPCReadSector(stream)
+	case rhp4.RPCRenewContractID:
+		err = s.handleRPCRenewContract(stream)
+	case rhp4.RPCSectorRootsID:
+		err = s.handleRPCSectorRoots(stream)
+	case rhp4.RPCWriteSectorID:
+		err = s.handleRPCWriteSector(stream)
+	case rhp4.RPCVerifySectorID:
+		err = s.handleRPCVerifySector(stream)
```
lukechampine, line 997: `map[types.Specifier]func(net.Conn) error`? :)

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+
+	// calculate the cost
+	cost := prices.RPCWriteSectorCost(req.DataLength, req.Duration)
+	// debit the account
+	if err := s.contractor.DebitAccount(req.Token.Account, cost); err != nil {
+		return fmt.Errorf("failed to debit account: %w", err)
+	}
+
+	// store the sector
+	root := rhp4.SectorRoot(&sector)
```
lukechampine, line 243: this should be fused with the `ReadFull` above (via `ReaderRoot`?) for performance

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+	var req rhp4.RPCLatestRevisionRequest
+	if err := rhp4.ReadRequest(stream, &req); err != nil {
+		return errorDecodingError("failed to read request: %v", err)
+	}
+
+	state, unlock, err := s.contractor.LockV2Contract(req.ContractID)
+	if err != nil {
+		return fmt.Errorf("failed to lock contract: %w", err)
+	}
+	defer unlock()
```
lukechampine, line 404: not a big deal, but presumably we can unlock as soon as we observe the `state.Revision`, instead of waiting until after we write the response?

In general we want to avoid holding locks during I/O. Then again, if it only matters when multiple renters are accessing the same contract in parallel, we probably don't need to worry about that.

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+			Parent: sce,
+		})
+		renterInputs = renterInputs.Add(sce.SiacoinOutput.Value)
+	}
+
+	// calculate the required funding
+	cs := s.chain.TipState()
+	renterCost, hostCost := rhp4.ContractCost(cs, prices, formationTxn.FileContracts[0], formationTxn.MinerFee)
+	// validate the renter added enough inputs
+	if n := renterInputs.Cmp(renterCost); n < 0 {
```
lukechampine, line 522: comparisons are cheap; might be clearer to write this as:
```go
if renterInputs.Cmp(renterCost) < 0 {

} else if renterInputs != renterCost {

}
```

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+	}
+
+	// TODO: build proof
+	resp := rhp4.RPCVerifySectorResponse{
+		Leaf: ([64]byte)(sector[rhp4.LeafSize*req.LeafIndex:]),
+	}
+	return rhp4.WriteResponse(stream, &resp)
+}
+
+func updateStateElementBasis(cm ChainManager, base, target types.ChainIndex, element *types.StateElement) error {
```
lukechampine, line 839: `ChainManager` has a method for this, right? But IIRC it expects a full `types.V2Transaction`. Seems to me like we should expose that somehow instead of defining `updateStateElementBasis` and `updateSiacoinElementBasis` here.

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+		rhp4.WriteResponse(stream, &rhp4.RPCError{Code: rhp4.ErrorCodeBadRequest, Description: "unrecognized RPC"})
+		return
+	}
+	if err != nil {
+		var re *rhp4.RPCError
+		if ok := errors.As(err, &re); ok {
+			rhp4.WriteResponse(stream, re)
+			log.Debug("RPC failed", zap.Error(err), zap.Duration("elapsed", time.Since(rpcStart)))
+		} else {
+			rhp4.WriteResponse(stream, rhp4.ErrHostInternalError.(*rhp4.RPCError))
```
lukechampine, line 1009: is the assert necessary here?

rhp/v4/rpc.go
```diff
@@ -0,0 +1,718 @@
...
+	TransactionSet struct {
+		Basis        types.ChainIndex      `json:"basis"`
+		Transactions []types.V2Transaction `json:"transactions"`
+	}
+
+	// ContractRevision pairs a contract ID with a revision.
+	ContractRevision struct {
+		ID       types.FileContractID `json:"id"`
+		Revision types.V2FileContract `json:"revision"`
+	}
```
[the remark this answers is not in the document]
n8maninger: We definitely want to avoid passing around the proof as much as possible.

rhp/v4/server.go
n8maninger: IIRC only `AddV2PoolTransaction` updates the proofs. The problem that you run into is the host and renter are passing elements around with proofs at various heights. I'm fine with moving this into the ChainManager. It feels like a missed opportunity that each state element can't communicate how up-to-date its proof is.

testutil/host.go
n8maninger: `hostd` is all `rhp2` and `rhp3` and `renterd` is a mix of both 😓

rhp/v4/rpc.go
n8maninger: Is this about the miner fee? It's totally about the miner fee isn't it?

rhp/v4/rpc.go
```diff
@@ -0,0 +1,718 @@
...
+
+type (
+	// A TransportClient is a generic multiplexer for outgoing streams.
+	TransportClient interface {
+		DialStream(context.Context) net.Conn
+		Close() error
+	}
+
+	// A ChainReader reports the chain state and manages the transaction pool.
+	ChainReader interface {
```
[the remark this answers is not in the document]
n8maninger: There's a `ChainManager` in server.go. I'd prefer to have them remain separate, but maybe I'm being overly pedentic.

chain/hostannouncement.go
```diff
@@ -11,36 +12,68 @@ const attestationHostAnnouncement = "HostAnnouncement"
...
+	HostAnnouncement struct {
+		PublicKey  types.PublicKey `json:"publicKey"`
+		NetAddress string          `json:"netAddress"`
+	}
+
+	// A Protocol is a string identifying a network protocol that a host may be
+	// reached on. It should be in the form of "protocol/transport". There may be
+	// additional useful information after the transport, separated by forward
+	// slashes.
+	Protocol string
```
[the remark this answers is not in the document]
n8maninger: - `rhp4/tcp/siamux`
- `rhp4/websocket`

It is probably overkill to worry about it being any more than an ordinary string

chain/hostannouncement.go
```diff
@@ -11,36 +12,68 @@ const attestationHostAnnouncement = "HostAnnouncement"
...
+func (ha *V2HostAnnouncement) FromAttestation(a types.Attestation) error {
 	if a.Key != attestationHostAnnouncement {
-		return false
+		return errors.New("not a host announcement")
 	}
-	ha.PublicKey = a.PublicKey
-	ha.NetAddress = string(a.Value)
-	return true
+	d := types.NewBufDecoder(a.Value)
+	types.DecodeSlice(d, (*[]NetAddress)(ha))
```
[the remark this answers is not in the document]
n8maninger: It does not

rhp/v4/server.go
n8maninger: `rhp4.ErrHostInternalError` is an `error` so no implementation of Object, but we could change that

rhp/v4/rpc.go
```diff
@@ -0,0 +1,718 @@
...
+	// RPCModifySectorsResult contains the result of executing the modify sectors RPC
+	RPCModifySectorsResult struct {
+		Revision types.V2FileContract `json:"revision"`
+		Cost     types.Currency       `json:"cost"`
+	}
+
+	// RPCFundAccountResult contains the result of executing the fund accounts RPC
+	RPCFundAccountResult struct {
+		Revision types.V2FileContract `json:"revision"`
+		Balances []AccountBalance     `json:"balances"`
```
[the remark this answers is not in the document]
n8maninger: Yeah, I prefer slices over maps for this purpose. More inline with how it is ideally consumed.

rhp/v4/rpc.go
lukechampine: it's about the miner fee

honestly 1 SC is a fine miner fee (well above average, probably) but hardcoding it...

testutil/host.go
lukechampine: rip

rhp/v4/server.go
lukechampine: ah right, gotcha. I think I was getting it mixed up with `rhpv2` which has a `WriteResponseErr`.

rhp/v4/server.go
lukechampine: Like `types.StateElement` would have a `basis ChainIndex` field? That feels excessive to me, given that in most contexts, every basis in a transaction/block will be identical. It's only relevant when communicating with peers -- which is why we make it explicit in RHPv4 (as well as `AddV2PoolTransaction`).

Seems like just exporting `(ChainManager).updateV2Transaction` would do the trick. Taking a step back, though: do we actually need to be updating the elements here? If the renter's elements aren't up-to-date, `AddV2PoolTransaction` will update them; if they're *too* outdated, it will reject them, and we'll refuse to send our signatures to the renter.

rhp/v4/rpc.go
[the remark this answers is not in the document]
lukechampine: ah, I see. What if we remove `Tip()` and `TipState()` -- those can be passed as arguments -- and call it `TxPool`?

rhp/v4/server.go
n8maninger: Will it not cause problems if we call Update on a transaction when the host's elements are `basis A` and the renter's elements are `basis B`?

rhp/v4/rpc.go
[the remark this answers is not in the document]
n8maninger: That sounds like a good solution

rhp/v4/rpc.go
n8maninger: Yeah, we should not do that 😅

chain/hostannouncement.go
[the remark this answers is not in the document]
lukechampine: bummer

rhp/v4/server.go
```diff
@@ -0,0 +1,1078 @@
...
+	"go.sia.tech/coreutils/wallet"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+// maxBasisDiff is the maximum number of blocks by which a transaction's basis
+// can differ from the current tip.
+const maxBasisDiff = 20
+
+var protocolVersion = [3]byte{0, 0, 1}
```
lukechampine, line 25: ...isn't this version 4? 😑

rhp/v4/server.go
n8maninger: I really dislike `map[types.Specifier]func(net.Conn) error`. It is probably fine in practice and is probably optimized away by Go's compiler, but it looks meh as a replacement for a switch statement when the number of cases is relatively small and known ahead of time. For huge switch statements it can be better for maintenance, but 10 or so cases isn't that.

n8maninger: To be clear: I don't mind changing it, but I will grumble about it 😛

rhp/v4/server.go
n8maninger: I will note that was a left over from the summit

rhp/v4/server.go
```diff
@@ -0,0 +1,1037 @@
...
+	if err := rhp4.ReadRequest(stream, &req); err != nil {
+		return errorDecodingError("failed to read request: %v", err)
+	}
+
+	if err := req.Validate(s.hostKey.PublicKey()); err != nil {
+		return errorBadRequest("request invalid: %v", err)
+	}
+	prices, token := req.Prices, req.Token
+
+	if err := s.contractor.DebitAccount(token.Account, prices.RPCReadSectorCost(req.Length)); err != nil {
```
ChrisSchinnerl, line 186: How do you feel about a quick check whether the host has the sector before we debit the account? For two reasons:
1. A renter will be able to find out whether a host has a sector without paying
2. A renter isn't charged full price for what is essentially a 404

Especially 1. is useful or maybe essential in RHP4 since clients don't necessarily know whether we have a contract with a host.

rhp/v4/server.go
lukechampine: yeah, fine as is -- I agree it's largely an aesthetic choice. Personally I like having a list of `name -> handler` mappings, one per line, but it *will* be less efficient than a `switch`.

rhp/v4/server.go
lukechampine: ah, I see. Yeah, you can't mix-and-match bases, all the elements have to have the same basis. So we need a method like:
```go
func (ChainManager) ChangeV2TransactionSetBasis(txns []types.V2Transaction, from, to types.ChainIndex) error
```
Then the process is:
1. Create a txn set with some basis
2. Receive elements from peer with some other basis
3. Change peer elements to your basis
4. Submit to txpool (which may do another update, if a block has been mined since you created the txn set)

n8maninger: We don't get the actual transaction from the peer in this case, but maybe we should just build a temp transaction so that it can be a bit more generic?

rhp/v4/rpc.go
```diff
@@ -0,0 +1,705 @@
...
+	} else if n != int64(resp.DataLength) {
+		return RPCReadSectorResult{}, io.ErrUnexpectedEOF
+	}
+	return RPCReadSectorResult{
+		Cost: prices.RPCReadSectorCost(length),
+	}, nil
+}
+
+// RPCWriteSector writes a sector to the host.
+func RPCWriteSector(ctx context.Context, t TransportClient, prices rhp4.HostPrices, token rhp4.AccountToken, rl ReaderLen, duration uint64) (RPCWriteSectorResult, error) {
```
peterjan, line 215: What's the benefit of the `ReaderLen` interface over passing in the amount of data to read?

rhp/v4/rpc.go
```diff
@@ -0,0 +1,705 @@
...
+
+	bw := bufio.NewWriterSize(stream, t.FrameSize())
+
+	if err := rhp4.WriteRequest(bw, rhp4.RPCWriteSectorID, &req); err != nil {
+		return RPCWriteSectorResult{}, fmt.Errorf("failed to write request: %w", err)
+	}
+
+	sr := io.LimitReader(rl, int64(req.DataLength))
+	tr := io.TeeReader(sr, bw)
+	if req.DataLength < rhp4.SectorSize {
```
peterjan, line 243: Should we treat the case where `DataLength` is 0 and we're writing an empty sector?

rhp/v4/rpc.go
```diff
@@ -308,8 +316,8 @@ func RPCModifySectors(ctx context.Context, t TransportClient, cs consensus.State
 	}
 
 	// TODO: verify proof
-
-	revision, err := rhp4.ReviseForModifySectors(contract.Revision, req, resp)
+	root := resp.Proof[len(resp.Proof)-1]
```
peterjan, line 321: we should validate the proof's not empty

n8maninger:
I think most of the contents has been reviewed already. One last run through for all of us and lets get this milestone merged.

rhp/v4/rpc.go
```diff
@@ -0,0 +1,910 @@
...
+	hostFormationTxn := hostTransactionSetResp.TransactionSet[len(hostTransactionSetResp.TransactionSet)-1]
+	if len(hostFormationTxn.FileContracts) != 1 {
+		return RPCFormContractResult{}, fmt.Errorf("expected exactly one contract")
+	}
+
+	// check for no funny business
+	formationTxnID := formationTxn.ID()
+	hostFormationTxnID := hostFormationTxn.ID()
+	if formationTxnID != hostFormationTxnID {
+		return RPCFormContractResult{}, errors.New("transaction ID mismatch")
```
ChrisSchinnerl, line 633: should've done `errors.New("funny business detected")`

rhp/v4/server.go
```diff
@@ -0,0 +1,1110 @@
...
+	settings := s.settings.RHP4Settings()
+	tip := s.chain.Tip()
+	if err := req.Validate(ourKey, tip, settings.MaxCollateral, settings.MaxContractDuration); err != nil {
+		return err
+	}
+	prices := req.Prices
+
+	fc, usage := rhp4.NewContract(prices, req.Contract, ourKey, settings.WalletAddress)
+	formationTxn := types.V2Transaction{
+		MinerFee:      req.MinerFee,
```
ChrisSchinnerl, line 530: I thought we talked about eliminating the fee from the request and having the host fund the fee in exchange for the renter funding the `ContractPrice`?
It doesn't matter too much but it feels like the host would have the better estimate of network fees considering that it is running 24/7

rhp/v4/rpc_test.go
```diff
@@ -0,0 +1,1561 @@
...
+	for i, n := range indices {
+		newRoots[n] = newRoots[len(newRoots)-i-1]
+	}
+	newRoots = newRoots[:len(newRoots)-len(indices)]
+
+	removeResult, err := rhp4.RPCFreeSectors(context.Background(), transport, cs, settings.Prices, renterKey, revision, indices)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assertRevision(t, removeResult.Revision, newRoots)
```
ChrisSchinnerl, line 1150: This is not the final behavior yet right? Since we wanted to replace the sectors with sentinel values and decrease the size but keep the capacity rather than actually remove the roots.

rhp/v4/server.go
n8maninger: Since the host only validates the fee is above 0 and the transaction is valid, it didn't seem necessary to change the fee payer too. The host will accept the contract and retry the transaction until it's confirmed or the reject height hits. If there's a period of high fees, the formation transaction will probably be accepted later.

ChrisSchinnerl (approved):

PR #102, opened 2024-10-03
rhp4,chain: simplify renew, move proof updates to chain manager

n8maninger:
Simplifies renew with https://github.com/SiaFoundation/core/pull/209 and moves proof updates into the chain manager

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -949,7 +949,7 @@ func (m *Manager) V2TransactionSet(basis types.ChainIndex, txn types.V2Transacti
 	m.revalidatePool()
 
 	// update the transaction's basis to match tip
-	_, txns, err := m.updateV2TransactionSet(basis, []types.V2Transaction{txn})
+	txns, err := m.UpdateV2TransactionSet([]types.V2Transaction{txn}, basis, m.tipState.Index)
 	if err != nil {
 		return types.ChainIndex{}, nil, fmt.Errorf("failed to update transaction set basis: %w", err)
 	} else if len(txns) == 0 {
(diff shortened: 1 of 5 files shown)


rhp/v4/server.go
```diff
@@ -708,61 +707,49 @@ func (s *Server) handleRPCRenewContract(stream net.Conn) error {
...
+		renewalTxn = updated[0]
+		renewalTxn.SiacoinInputs = append(renewalTxn.SiacoinInputs, hostInputs...)
 	}
-	// update the file contract element to reflect the funding basis
-	if fceBasis != basis {
-		if err := updateStateElementBasis(s.chain, fceBasis, basis, &fce.StateElement); err != nil {
-			return errorBadRequest("failed to update file contract basis: %v", err)
+
+	if elementBasis != basis {
+		if err := s.chain.UpdateStateElement(&fce.StateElement, elementBasis, basis); err != nil {
```
lukechampine, line 745: could we hack around the need for `UpdateStateElement` this by putting the contract in a transaction and using `UpdateV2TransactionSet`?

n8maninger: We could, just felt awkward. The duplicated logic is lame as well.

chain/manager.go
```diff
@@ -1086,47 +1086,86 @@ func (m *Manager) AddPoolTransactions(txns []types.Transaction) (known bool, err
...
 }
 
-// updateV2TransactionSet updates the basis of a transaction set to the current
-// tip. If the basis is already the tip, the transaction set is returned as-is.
-// Any transactions that were confirmed are removed from the set. Any ephemeral
-// state elements that were created by an update are updated.
+// UpdateStateElement updates the basis of a state element from "from" to "to".
+// If from and to are equal, the state element is not modified.
+func (m *Manager) UpdateStateElement(se *types.StateElement, from, to types.ChainIndex) error {
+	if from == to {
```
lukechampine, line 1092: needs a lock; which in turn means `AddV2PoolTransactions` needs to avoid double-locking

rhp/v4/server.go
```diff
@@ -708,61 +707,49 @@ func (s *Server) handleRPCRenewContract(stream net.Conn) error {
...
+			Value:   renterInputSum.Sub(renterCost),
+		})
 	}
 
-	fceBasis, fce, err := s.contractor.ContractElement(req.Renewal.ContractID)
+	elementBasis, fce, err := s.contractor.ContractElement(req.Renewal.ContractID)
 	if err != nil {
 		return fmt.Errorf("failed to get contract element: %w", err)
 	}
-	renewalTxn.FileContractResolutions[0].Parent = fce
```
peterjan, line 720: We no longer set the element as parent on the resolution in the renewal txn here, while `UpdateV2TransactionSet` calls `updateTxnProofs` which includes a step where update the proof of the resolutions `updateProof(&txn.FileContractResolutions`.. is that an oversight or is that intentional?

peterjan: Oh I guess `UpdateStateElement` covers that

chain/manager.go
peterjan: Could you elaborate on why it needs a lock? I don't get it.

rhp/v4/rpc.go
```diff
@@ -649,48 +649,28 @@ func RPCRenewContract(ctx context.Context, t TransportClient, tp TxPool, signer
...
-
-		renewalTxn.SiacoinInputs = append(renewalTxn.SiacoinInputs, types.V2SiacoinInput{
-			Parent: setupTxn.EphemeralSiacoinOutput(0),
-		})
-		signer.SignV2Inputs(&renewalTxn, []int{0})
+	req := rhp4.RPCRenewContractRequest{
+		Prices:   p,
+		Renewal:  params,
+		MinerFee: renewalTxn.MinerFee,
+		Basis:    cs.Index,
```
peterjan, line 656: `nit` no need to set the basis as we override it anyway

chain/manager.go
lukechampine: I mean this method as a whole; we need to ensure a consistent view of the blockchain across this function

rhp/v4/server.go
n8maninger: That's intentional as we don't want the chain manager to try and update the zero value proof.

peterjan (approved):

PR #104, opened 2024-10-07
rhp4 RPC Refresh

n8maninger:
Implements RPC refresh for more efficient contract refreshes 

Requires #102 and https://github.com/SiaFoundation/core/pull/212


rhp/v4/rpc.go
```diff
@@ -754,3 +761,131 @@ func RPCRenewContract(ctx context.Context, t TransportClient, tp TxPool, signer
 		Cost: renterCost,
 	}, nil
 }
+
+// RPCRefreshContract renews a contract with a host.
```
peterjan, line 765:
```suggestion
// RPCRefreshContract refreshes a contract with a host.
```

rhp/v4/rpc_test.go
```diff
@@ -378,70 +377,119 @@ func TestRenewContractPartialRollover(t *testing.T) {
...
+
+		// renew the contract
+		_, err = rhp4.RPCRefreshContract(context.Background(), transport, cm, fundAndSign, cm.TipState(), settings.Prices, revision.Revision, proto4.RPCRefreshContractParams{
+			ContractID: revision.ID,
+			Allowance:  types.ZeroCurrency,
+			Collateral: types.ZeroCurrency,
+		})
+		if err == nil {
+			t.Fatal(err)
+		} else if !strings.Contains(err.Error(), "allowance is zero") {
```
peterjan, line 425: could a case for collateral being greater than max collateral

rhp/v4/rpc_test.go
```diff
@@ -378,70 +377,119 @@ func TestRenewContractPartialRollover(t *testing.T) {
...
-		Collateral:  types.Siacoins(300),
-		ProofHeight: revision.Revision.ProofHeight + 10,
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
+	t.Run("valid refresh", func(t *testing.T) {
+		fundAndSign := &fundAndSign{w, renterKey}
+		renterAllowance, hostCollateral := types.Siacoins(100), types.Siacoins(200)
+		result, err := rhp4.RPCFormContract(context.Background(), transport, cm, fundAndSign, cm.TipState(), settings.Prices, hostKey.PublicKey(), settings.WalletAddress, proto4.RPCFormContractParams{
```
peterjan, line 433: extracting these ~40 lines into a test helper that forms a contract and verifies the transaction set is valid, perhaps also funds an account would save a lot of lines in this file, seems like these steps are repeated 5-6 times over and are exact copies

rhp/v4/server.go
```diff
@@ -669,29 +669,193 @@ func (s *Server) handleRPCRenewContract(stream net.Conn) error {
...
+
+	if elementBasis != basis {
+		tempTxn := types.V2Transaction{
+			FileContractResolutions: []types.V2FileContractResolution{
+				{Parent: fce, Resolution: &renewal},
+			},
+		}
+		updated, err := s.chain.UpdateV2TransactionSet([]types.V2Transaction{tempTxn}, elementBasis, basis)
+		if err != nil {
+			return fmt.Errorf("failed to update contract element: %w", err)
```
peterjan, line 749: If we return a bad request when we update the renter inputs to reflect the chain state, shouldn't we return one here as well?

peterjan (approved):
Nothin' but nits and small comments, LGTM 👍

rhp/v4/server.go
n8maninger: Bad request should only be for issues with parameters from the renter. This error should be a host error since all of the input parameters are from the host.

PR #116, opened 2024-11-09
Add support for block pruning

lukechampine:
Been a long time coming. 😅 

The strategy here is quite naive, but I think it will be serviceable. Basically, when we apply a block `N`, we delete block `N-P`. `P` is therefore the "prune target," i.e. the maximum number of blocks you want to store.

In practice, this isn't exhaustive: it only deletes blocks from the best chain. It also won't dramatically shrink the size of an existing database. I think this is acceptable, because pruning is most important during the initial sync, and during the initial sync, you'll only be receiving blocks from one chain at a time. Also, we don't want to make pruning *too* easy; after all, we need a good percentage of nodes to be storing the full chain, so that others can sync to them.

I tested this out locally with a prune target of 1000, and after syncing 400,000 blocks, my `consensus.db` was around 18 GB. This is disappointing; it should be much smaller. With some investigation, I found that the Bolt database was only storing ~5 GB of data (most of which was the accumulator tree, which we can't prune until after v2). I think this is a combination of a) Bolt grows the DB capacity aggressively in response to writes, and b) Bolt never shrinks the DB capacity. So it's possible that we could reduce this number by tweaking our DB batching parameters. Alternatively, we could provide a tool that copies the DB to a new file. Not the most user-friendly, but again, I think I'm okay with that for now.

Depends on https://github.com/SiaFoundation/core/pull/228

### DIFF
--- a/.changeset/add_support_for_block_pruning.md
+++ b/.changeset/add_support_for_block_pruning.md
@@ -1,0 +1,17 @@
+---
+default: minor
+---
+
+# Add support for block pruning
+
+#116 by @lukechampine
+
+Been a long time coming. 😅 
+
+The strategy here is quite naive, but I think it will be serviceable. Basically, when we apply a block `N`, we delete block `N-P`. `P` is therefore the "prune target," i.e. the maximum number of blocks you want to store.
+
+In practice, this isn't exhaustive: it only deletes blocks from the best chain. It also won't dramatically shrink the size of an existing database. I think this is acceptable, because pruning is most important during the initial sync, and during the initial sync, you'll only be receiving blocks from one chain at a time. Also, we don't want to make pruning *too* easy; after all, we need a good percentage of nodes to be storing the full chain, so that others can sync to them.
+
+I tested this out locally with a prune target of 1000, and after syncing 400,000 blocks, my `consensus.db` was around 18 GB. This is disappointing; it should be much smaller. With some investigation, I found that the Bolt database was only storing ~5 GB of data (most of which was the accumulator tree, which we can't prune until after v2). I think this is a combination of a) Bolt grows the DB capacity aggressively in response to writes, and b) Bolt never shrinks the DB capacity. So it's possible that we could reduce this number by tweaking our DB batching parameters. Alternatively, we could provide a tool that copies the DB to a new file. Not the most user-friendly, but again, I think I'm okay with that for now.
+
+Depends on https://github.com/SiaFoundation/core/pull/228
--- a/.changeset/add_support_for_block_pruning.md
+++ b/.changeset/add_support_for_block_pruning.md
@@ -4,14 +4,4 @@ default: minor
 
 # Add support for block pruning
 
-#116 by @lukechampine
-
-Been a long time coming. 😅 
-
-The strategy here is quite naive, but I think it will be serviceable. Basically, when we apply a block `N`, we delete block `N-P`. `P` is therefore the "prune target," i.e. the maximum number of blocks you want to store.
-
-In practice, this isn't exhaustive: it only deletes blocks from the best chain. It also won't dramatically shrink the size of an existing database. I think this is acceptable, because pruning is most important during the initial sync, and during the initial sync, you'll only be receiving blocks from one chain at a time. Also, we don't want to make pruning *too* easy; after all, we need a good percentage of nodes to be storing the full chain, so that others can sync to them.
-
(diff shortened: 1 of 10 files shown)


chain/db.go
```diff
@@ -13,43 +13,32 @@ import (
 )
 
 type supplementedBlock struct {
-	Block      types.Block
+	Header     *types.BlockHeader
```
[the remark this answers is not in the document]
lukechampine: agreed. In this case, though, compatibility is preserved: we always encode as v3, but we can decode either v2 or v3. That's possible because the `Header` field is optional and gets filled in lazily as needed.

chain/manager.go
```diff
@@ -314,6 +318,13 @@ func (m *Manager) applyTip(index types.ChainIndex) error {
 	m.store.ApplyBlock(cs, cau)
 	m.applyPoolUpdate(cau, cs)
 	m.tipState = cs
+
+	if m.pruneTarget != 0 && cs.Index.Height > m.pruneTarget {
+		//m.log.Info("pruning block", zap.Uint64("height", cs.Index.Height-m.pruneTarget))
```
n8maninger, line 323: You can leave this in with `.Debug`

testutil/syncer.go
```diff
@@ -7,15 +7,15 @@ import (
 	"go.sia.tech/coreutils/syncer"
 )
 
-// A MemPeerStore is an in-memory implementation of a PeerStore.
-type MemPeerStore struct {
+// A EphemeralPeerStore is an in-memory implementation of a PeerStore.
+type EphemeralPeerStore struct {
```
ChrisSchinnerl, line 11: Not sure how strict we want to be just yet but `coreutils` is a library and this is breaking part of the libraries interface which makes it a breaking change and not a `minor`.

ChrisSchinnerl (commented):
This change is pretty scary. It seems like pruning is now done but I think the way it is now isn't safe for the network.

We now have a state where users might prune blocks without the syncer notifying other peers about not being full node. So imo this is only safe to use once we have a solution for that.

Imo that involves signaling whether you are a full node and what your latest known block is. Then have the syncer only connect to peers which are either full nodes or "partial" nodes >= your required block height. Otherwise we might run into weird issues and network fragmentation in the worst case.

.changeset/add_support_for_block_pruning.md
```diff
@@ -0,0 +1,17 @@
+---
+default: minor
+---
+
+# Add support for block pruning
+
+#116 by @lukechampine
```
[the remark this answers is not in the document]
n8maninger: It could probably be condensed a bit more, but I like the additional context on big changes.

testutil/syncer.go
n8maninger: Breaking would push this to v1.0.0, which I don't think we're ready for quite yet.

.changeset/add_support_for_block_pruning.md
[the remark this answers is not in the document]
ChrisSchinnerl: Ahhhhh I see. Ok then it's fine. But I'd condense the body a bit. Make it sound a bit more neutral, no emojis etc. More like a wiki and less like an internal discussion.

testutil/syncer.go
lukechampine: just me being pedantic about consistent names. Probably not worth the breakage

lukechampine:
We should definitely test what happens when you try to sync from a pruned node. Looking at `BlocksForHistory`, I think the RPC will just fail, and you'll switch to a different peer. Seems fine to me.

n8maninger:
> RPC will just fail, and you'll switch to a different peer

Will the pruned peer ever be dinged enough to be blocked? That could cause problems if the pruned node ends up blocked by all the full nodes over time 🤔

lukechampine:
Currently `syncLoop` doesn't penalize failed syncs at all, it just logs the error.

.changeset/add_support_for_block_pruning.md
```diff
@@ -0,0 +1,17 @@
...
+
+Been a long time coming. 😅
```
n8maninger, line 9: @lukechampine did you want to clean this up a bit before merging?

n8maninger (dismissed):

.changeset/add_support_for_block_pruning.md
n8maninger: It didn’t have to be that dry😅

PR #120, opened 2024-11-22
chain: Handle txpool conflicts properly

lukechampine:
This PR changes the txpool to reject transactions that conflict with transactions already in the pool. As a result, the set of transactions in the pool should always be mutually valid (albeit possibly too large to fit in one block). Previously, conflicting transactions were allowed to coexist; `siad` does this intentionally, but in `coreutils` it was accidental, leading to surprising and confusing behavior.

Note that there is currently no "replace-by-fee" mechanism for clearing stuck transactions from the pool. This is a somewhat contentious topic, so I'm punting on it for now. Longer term, I expect we'll want at least a basic child-pays-for-parent scheme; see the comment at the top of `revalidatePool`.

Also fixed a bug in `revalidatePool` where low-fee v2 transactions were not removed.

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -473,7 +473,7 @@ func (m *Manager) revalidatePool() {
 	// transactions, but that's expensive; this approach should work fine in
 	// practice.
 	if m.txpool.weight >= txpoolMaxWeight {
-		// sort txns fee without modifying the actual pool slice
+		// sort txns by fee without modifying the actual pool slice
 		type feeTxn struct {
 			index  int
 			fees   types.Currency
@@ -507,10 +507,16 @@ func (m *Manager) revalidatePool() {
 			return txnFees[i].index < txnFees[j].index
 		})
 		rem := m.txpool.txns[:0]
+		v2rem := m.txpool.v2txns[:0]
 		for _, ft := range txnFees {
-			rem = append(rem, m.txpool.txns[ft.index])
+			if !ft.v2 {
+				rem = append(rem, m.txpool.txns[ft.index])
+			} else {
+				v2rem = append(v2rem, m.txpool.v2txns[ft.index])
+			}
 		}
 		m.txpool.txns = rem
+		m.txpool.v2txns = v2rem
 	}
 
 	// remove and re-add all transactions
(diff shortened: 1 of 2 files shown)


wallet/wallet_test.go
```diff
@@ -1946,11 +1946,16 @@ func TestV2TPoolRace(t *testing.T) {
...
 
-	// broadcast the transaction set including the already confirmed setup
-	// transaction. This seems unnecessary, but it's a fairly common occurrence
-	// when passing transaction sets using unconfirmed outputs between a renter
-	// and host. If the transaction set is not updated correctly, it will fail.
-	if _, err := cm.AddV2PoolTransactions(basis, []types.V2Transaction{setupTxn, spendTxn}); err != nil {
+	// now that the setup transaction has been accepted, we can't include it in
+	// the set anymore, or we'll get a double-spend error. AddV2PoolTransactions
+	// will internally replace the ephemeral output, making spendTxn valid.
+	if _, err := cm.AddV2PoolTransactions(basis, []types.V2Transaction{spendTxn}); err != nil {
```
n8maninger, line 1952: Should extend this with a case that broadcasts both and checks the returned error for a double spend.

ChrisSchinnerl: +1

chain/manager.go
```diff
@@ -1001,44 +1007,56 @@ func (m *Manager) V2TransactionSet(basis types.ChainIndex, txn types.V2Transacti
...
+func (m *Manager) checkTxnSet(txns []types.Transaction, v2txns []types.V2Transaction) (bool, error) {
 	allInPool := true
-	checkPool := func(txid types.TransactionID) {
+	checkPool := func(txid types.TransactionID) types.TransactionID {
 		if allInPool {
 			if _, ok := m.txpool.indices[txid]; !ok {
 				allInPool = false
 			}
 		}
+		return txid
```
ChrisSchinnerl, line 1018: Kinda surprised you didn't go all the way to save those extra 3 lines of code :P 
```go
_, ok := m.txpool.indices[txid]
allInPool = allInPool && ok
```

chain/manager.go
```diff
@@ -1001,44 +1007,56 @@ func (m *Manager) V2TransactionSet(basis types.ChainIndex, txn types.V2Transacti
...
 	}
-	for i, txn := range v2txns {
-		txid := txn.ID()
-		checkPool(txid)
-		h.E.WriteUint64(uint64(i))
-		txid.EncodeTo(h.E)
+	for _, txn := range v2txns {
+		checkPool(txn.ID()).EncodeTo(h.E)
 	}
 	setID := h.Sum()
```
ChrisSchinnerl, line 1027: Would there be any benefit in making the order of the transactions within the set deterministic to make sure the order doesn't influence the set id?

chain/manager.go
lukechampine: I would, but Go isn't smart enough to skip the map lookup if `allInPool` is already `false`!

Same thing applies to `for` loops:
```go
var all bool
for _, x := range xs {
    all = all || x
}
```
In theory the compiler could break out of the loop as soon as `all = true`, but it doesn't; you have to write it yourself. Sad.

chain/manager.go
lukechampine: pretty small, I think; most of the time, transactions can't be reordered anyway due to dependencies.

(In fact, I wonder if we should *reject* transaction sets that *aren't* internally dependent...hmm. Probably not worth it.)

wallet/wallet_test.go
lukechampine: huh, apparently `AddV2PoolTransactions` will also silently skip any transactions that were confirmed since `basis`. Definitely convenient, but it does feel a bit weird that we don't do it for v1 transactions too. 🤔

n8maninger: That behavior is useful for rebroadcasting contract formation sets on the host

n8maninger: Avoiding the race is a pretty big benefit of keeping them combined. Basis and proof updates in RHP4 have enough complexity without adding another bit 😓 

I'm not sure if updating `AddPoolTransactions` to include a basis is worth it. v1 isn't equipped to pass it around, unlike v2 where we explicitly store basis alongside the transaction set.

PR #121, opened 2024-11-22
Add missing account-related errors for rhp4

ChrisSchinnerl:
The renter and host need to agree on errors to return for individual RPCs. This PR adds the expected behavior to the corresponding interface methods.

Right now hostd returns an internal error when an account doesn't exist. That's suboptimal since renterd can't take action based on that. Returning an error for missing accounts seems a bit redundant so I propose just returning a balance of 0 similar to how we do in rhp3.


ChrisSchinnerl:
Closing this since it probably belongs into `core`.

PR #122, opened 2024-11-26
Return `V2FileContractRevision` in `RPCLatsetRevision` rather than `V2FileContract`

ChrisSchinnerl:
Ran into a little blocker on the rhp4 revision broadcast PR https://github.com/SiaFoundation/renterd/pull/1698

Basically I can't broadcast the revision since the renter doesn't keep track of every revision and the host doesn't return the state element required to broadcast one.

This PR is obviously incomplete without the corresponding `core` update but before opening that as well I wanted to draw attention to the issue.

Essentially this PR changes the `Contract` field on the response to `Revision` and adds a `ChainIndex` as well.

### DIFF
--- a/rhp/v4/rpc.go
+++ b/rhp/v4/rpc.go
@@ -474,11 +474,11 @@ func RPCFundAccounts(ctx context.Context, t TransportClient, cs consensus.State,
 }
 
 // RPCLatestRevision returns the latest revision of a contract.
-func RPCLatestRevision(ctx context.Context, t TransportClient, contractID types.FileContractID) (types.V2FileContract, error) {
+func RPCLatestRevision(ctx context.Context, t TransportClient, contractID types.FileContractID) (types.V2FileContractRevision, error) {
 	req := rhp4.RPCLatestRevisionRequest{ContractID: contractID}
 	var resp rhp4.RPCLatestRevisionResponse
 	err := callSingleRoundtripRPC(ctx, t, rhp4.RPCLatestRevisionID, &req, &resp)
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl:
> Why not track and update the state element in the renter database? Relying on the host seems counterintuitive. The renter knows the ID during formation. During consensus updates: loop through the v2 file contract elements, add any confirmed or revised v2 state elements, then update the proofs.

Mostly because it is very consistent with what we already have in the protocol. Where the renter doesn't need to know anything about revisions and the host provides all the information (e.g. proofs) to revise contracts. 

Storing works as well of course. Just with the minor downside of only the participants of the contract formation being the ones who can broadcast a revision versus anyone.

n8maninger:
> Where the renter doesn't need to know anything about revisions and the host provides all the information (e.g. proofs) to revise contracts.

The renter tracks contract status in the chain indexing, adding state elements to that is not that big of a change. Relying on the host for Merkle proofs means the renter won't be able to get their funds back if the host is offline at expiration.

> Just with the minor downside of only the participants of the contract formation being the ones who can broadcast a revision versus anyone.

That's not really different from how it works today.

lukechampine:
I am sympathetic, but I think I side with Nate. Including the proof seems to make life easier for lightweight `us`-style renters -- but lightweight renters necessarily need to rely on 3rd party services like `shard` and `muse`. In that vein, I imagine that there will be a 3rd party service (perhaps the explorer?) that provides the Merkle proof for any requested element.

PR #123, opened 2024-12-02
Update core

ChrisSchinnerl:


rhp/v4/server.go
```diff
@@ -253,7 +252,7 @@ func (s *Server) handleRPCWriteSector(stream net.Conn) error {
 		return errorDecodingError("failed to read sector data: %v", err)
 	}
 
-	usage := prices.RPCWriteSectorCost(req.DataLength, req.Duration)
+	usage := prices.RPCWriteSectorCost(req.DataLength)
```
ChrisSchinnerl, line 255: This is slightly odd to me. The `RPCWriteSectorCost` which is in `core` uses `TempSectorDuration` now but the `RPCWriteSectorRequest` which is also declared in `core`, still allows for specifying a `duration`. 
Is that a leftover or do we want to keep the protocol flexible even if we expect to always use `TempSectorDuration` @n8maninger ?

rhp/v4/rpc_test.go
```diff
@@ -251,6 +247,15 @@ func TestFormContract(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// validate contract signatures
```
ChrisSchinnerl, line 250: @n8maninger I added this because it seems like the RPC result didn't contain the renter signature. Which led to the host persisting the contract with an invalid signature as well and `RPCLatestRevision` returning the revision without signature. Causing the renter to be unable to broadcast the revision since it wasn't signed.

rhp/v4/server.go
n8maninger: That is left over and should be removed

n8maninger:
closing in favor of #126

PR #125, opened 2024-12-02
chain: Add fallback supplementBlock decoding

lukechampine:
Not sure how to test this. Do we have a corrupted consensus.db anywhere that we can use?


n8maninger:
If you download `walletd` v0.8.0 and sync Zen you should be able to get one pretty easily.

lukechampine:
ok, this is working. The breakage was due to https://github.com/SiaFoundation/core/pull/199, which changed the encoding of all state elements. Mercifully, although there were other encoding changes, they only affect v2 types, so they don't require compat code.

n8maninger:
Would it be better to do an explicit upgrade rather than to fallback on failure? It would slow down startup, but I'm a little concerned that fallbacks will accumulate and eventually get lost.

lukechampine:
hmm, yeah, you're right, don't want old encodings to stick around forever. I'll knock that out tomorrow.

lukechampine:
Closed in favor of #132

PR #132, opened 2024-12-09
feat(chain): Implement DB migration

lukechampine:
This adds a `MigrateDB` function that converts a DB from one version to the next. Accordingly, the per-block and per-state versions have been removed.

I've tested this against `walletd` v0.8.0, though not exhaustively.

Currently there is no live feedback during migration. We could add logging or some other form of callback. We also discussed calling `MigrateDB` inside `NewDBStore`; I'm open to that, but it would require changing the function signature if we want logs as well, so we'd need to update a lot of callsites.

### DIFF
--- a/chain/db.go
+++ b/chain/db.go
@@ -187,7 +187,7 @@ func (b *dbBucket) getRaw(key []byte) []byte {
 }
 
 func (b *dbBucket) get(key []byte, v types.DecoderFrom) bool {
-	val := b.getRaw(key)
+	val := b.getRaw(b.db.vkey(key))
 	if val == nil {
 		return false
 	}
@@ -210,11 +210,15 @@ func (b *dbBucket) put(key []byte, v types.EncoderTo) {
 	b.db.enc.Reset(&buf)
 	v.EncodeTo(&b.db.enc)
 	b.db.enc.Flush()
-	b.putRaw(key, buf.Bytes())
+	b.putRaw(b.db.vkey(key), buf.Bytes())
+}
+
+func (b *dbBucket) deleteRaw(key []byte) {
+	check(b.b.Delete(key))
 }
 
 func (b *dbBucket) delete(key []byte) {
-	check(b.b.Delete(key))
+	b.deleteRaw(b.db.vkey(key))
 }
 
 var (
@@ -232,12 +236,19 @@ var (
 
 // DBStore implements Store using a key-value database.
 type DBStore struct {
-	db  DB
-	n   *consensus.Network // for getState
-	enc types.Encoder
+	db      DB
+	n       *consensus.Network // for getState
+	version uint8
+	keyBuf  []byte
+	enc     types.Encoder
 
 	unflushed int
 	lastFlush time.Time
+}
+
+func (db *DBStore) vkey(key []byte) []byte {
+	db.keyBuf = append(db.keyBuf[:0], key...)
+	return append(db.keyBuf, db.version)
 }
 
 func (db *DBStore) bucket(name []byte) *dbBucket {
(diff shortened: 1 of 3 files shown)


chain/migrate.go
```diff
@@ -0,0 +1,159 @@
...
+		var fce types.FileContractElement
+		for _, key := range db.BucketKeys(bFileContractElements) {
+			if len(key) == 32 {
+				rewrite(bFileContractElements, key, (*oldFileContractElement)(&fce), &fce)
+			}
+		}
+		if err != nil {
+			return err
+		}
+		dbs.bucket(bVersion).putRaw(bVersion, []byte{2})
```
ChrisSchinnerl, line 150: Doesn't this break when it gets interrupted? Since it's not atomic?

chain/migrate.go
```diff
@@ -0,0 +1,159 @@
...
+	if v := d.ReadUint8(); v != 2 {
+		d.SetErr(fmt.Errorf("incompatible version (%d)", v))
+	}
+	(*consensus.State)(vs).DecodeFrom(d)
+}
+
+// MigrateDB upgrades the database to the latest version.
+func MigrateDB(db DB, n *consensus.Network) error {
+	if db.Bucket(bVersion) == nil {
+		return nil // nothing to migrate
```
ChrisSchinnerl, line 91: shouldn't we try to migrate from "no version" to version 2? rather than never migrating?

lukechampine: turns out we've always had a version lol, I just forgot https://github.com/SiaFoundation/coreutils/blob/a28e868d07995cc1a917b3bd75e6645653c75901/chain/db.go#L780

chain/migrate.go
lukechampine: yeah, not really sure how best to handle that. One thing I tried was "if a value decodes successfully into the new type, then it's already been converted so you can skip it." But that turns out to have false positives: that is, there are encodings that can be decoded into either the old type or the new type without causing an error.

we can control when flushes occur, so maybe the best we can do is store a "progress" entry every time we flush, e.g. `1 5912` would indicate that we've updated all blocks and 5912 states. That implies that `BucketKeys` must return the keys in the same order every time, though.

n8maninger: Best to make it atomic instead of trying to hack around atomicity.

n8maninger: That might cause problems though. IIRC, bbolt has a max transaction size.

lukechampine: right, a single transaction isn't gonna be viable if you're re-encoding every block.

tbh... this kinda is an argument for sticking a version byte in front of every entry 😬

n8maninger: Or a saner database 😛

lukechampine:
Ok, new approach. Every key is suffixed with a version. (Prefix would also work fine, and might be better for sorting...hmm)
When we migrate, we check the version byte on every key to determine whether it needs updating.
The old DB keys don't have a version byte at all, though, so instead we check the key length. This is feasible because all the keys have a fixed length, unlike values which can vary in size. This is also why the version is on the key, not the value: if it was on the value, then migrating the old DB (which lacks version bytes) would be ambiguous: you wouldn't be able to tell if the byte was actually a version byte, or part of an old DB value.

I tested this manually by migrating a consensus.db, loading it, and reading from it. When I migrate it again, it's a no-op. When migration is interrupted and resumed, the already-migrated keys are skipped as intended.

chain/db.go
```diff
@@ -757,13 +715,12 @@ func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBSto
...
-		n:  n,
+		db:      db,
+		n:       n,
+		version: 2,
 	}
-
-	// if the db is empty, initialize it; otherwise, check that the genesis
-	// block is correct
-	if dbGenesis, ok := dbs.BestIndex(0); !ok {
+	if version := dbs.bucket(bVersion).getRaw(dbs.vkey(bVersion)); len(version) != 1 {
```
n8maninger, line 722: Looks like this is currently panicking when trying to migrate a `walletd` v0.8.0 database: `failed to create chain store: panic during database initialization: bucket already exists`

lukechampine: hmm, steps to reproduce? I don't see that error; here's what I did:
- `git checkout v0.8.0 && go install ./cmd/walletd`
- `walletd -network zen`
- Wait for a few thousand blocks, then Ctrl-C
- In a test function, call `coreutils.OpenBoltChainDB`, then `chain.MigrateDB`, then `chain.NewDBStore`

n8maninger: I did basically the same, but checked out walletd master with coreutils replaced. Let me double check that I didn’t screw up.

lukechampine:
Closing this for now. Giving that `renterd` and `hostd` users will need to delete their consensus.db and re-sync anyway, it seems reasonable to ask that of `walletd` users as well. I do think this code represents a decent approach to the update process, so I'll keep it around on a local branch if we decide we need something like it later on.

PR #154, opened 2025-01-17
syncer: Return error from Connect if Syncer is closed

### LINKED ISSUE #100: Fix deadlock in syncer
A random deadlock that prevents our syncer from shutting down.            

https://github.com/SiaFoundation/renterd/actions/runs/11030847304/job/30636343643?pr=1574

[stacktrace.txt](https://github.com/user-attachments/files/17144377/stacktrace.txt)

_Originally posted by @peterjan in https://github.com/SiaFoundation/coreutils/issues/92#issuecomment-2376111458_

lukechampine:
Fixes #100


ChrisSchinnerl:
Looking good. Any chance other exposed methods would cause similar issues? Or is it just connect since it creates a new outgoing connection?

syncer/syncer.go
```diff
@@ -681,13 +681,24 @@ func (s *Syncer) Run(ctx context.Context) error {
...
+	s.mu.Unlock()
 	s.wg.Wait()
+	s.mu.Lock()
+	s.l = nil
 	return err
 }
 
 // Connect forms an outbound connection to a peer.
 func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
+	if s.l == nil {
```
peterjan, line 699: you need to acquire the lock to avoid a race

syncer/syncer.go
```diff
@@ -681,13 +681,24 @@ func (s *Syncer) Run(ctx context.Context) error {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.l == nil {
+		return nil // already closed
+	}
 	err := s.l.Close()
+	s.mu.Unlock()
 	s.wg.Wait()
+	s.mu.Lock()
+	s.l = nil
```
peterjan, line 693: `Addr()` exposes the listener's address so we have to add an extra `nil` check there as well

syncer/syncer.go
```diff
@@ -681,13 +681,24 @@ func (s *Syncer) Run(ctx context.Context) error {
 
 // Close closes the Syncer's net.Listener.
 func (s *Syncer) Close() error {
+	s.mu.Lock()
```
peterjan, line 684: Is it crazy to derive a cancel from the context we get passed in run and call that here? That would make this a lot more robust.

n8maninger: Absolutely crazy to pass contexts around outside of their scope

ChrisSchinnerl: I agree that passing contexts around outside of their "context" is not great. So I'd leave it the way it is for now and consider a follow up with a dedicated context or thread group for the `Syncer`.

syncer/syncer.go
peterjan: (I'd also move it under the mutex in the syncer struct so it's clear)

syncer/syncer.go
```diff
@@ -368,8 +371,8 @@ func (s *Syncer) relayV2TransactionSet(index types.ChainIndex, txns []types.V2Tr
 func (s *Syncer) allowConnect(ctx context.Context, peer string, inbound bool) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if s.l == nil {
```
ChrisSchinnerl, line 371: This was never happening apparently

ChrisSchinnerl (commented):
Changed it to use a dedicated `closed` boolean to avoid accidentally causing panics when dereferencing nil pointers. Also added a quick test to confirm we receive the right error when calling `Connect`.

peterjan (approved):

Issue #160, opened 2025-01-22
Syncing gets stuck on Zen


mike76-dev:
Reported by at least 2 people (not counting myself):
On Zen, syncing may randomly stop before reaching the current height. After analyzing the `peers.json`, it looks like one of the connected peers is not synced, and after syncing to that peer the `syncer` doesn't attempt to sync further.

n8maninger:
What app and version are the using?

n8maninger:
You should update the coreutils version for your node. `walletd v0.8.0` and `coreutils v0.3.3` have the v2 activation height at `100000`.

PR #161, opened 2025-01-22
Add TestSyncerChain

ChrisSchinnerl:
This PR adds a test to verify 3 syncers will all sync up to the same height when connected in a chain

Related to https://github.com/SiaFoundation/coreutils/issues/160

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -558,12 +558,16 @@ func (s *Syncer) syncLoop(ctx context.Context) error {
 		s.mu.Lock()
 		defer s.mu.Unlock()
 		for _, p := range s.peers {
-			if p.Synced() {
-				continue
-			}
 			peers = append(peers, p)
 		}
-		sort.Slice(peers, func(i, j int) bool {
+		// shuffle peers to make sure we don't always pick the same 3 peers in
+		// case they were all already synced and support v2. Also use stable sort to
+		// keep the shuffled order mostly intact
+		frand.Shuffle(len(peers), func(i, j int) { peers[i], peers[j] = peers[j], peers[i] })
+		sort.SliceStable(peers, func(i, j int) bool {
+			if peers[i].Synced() != peers[j].Synced() {
+				return !peers[i].Synced() // prefer unsynced peers
+			}
 			return peers[i].t.SupportsV2() && !peers[j].t.SupportsV2()
 		})
 		if len(peers) > 3 {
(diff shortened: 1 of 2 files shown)


syncer/syncer.go
```diff
@@ -558,12 +558,12 @@ func (s *Syncer) syncLoop(ctx context.Context) error {
 		s.mu.Lock()
 		defer s.mu.Unlock()
 		for _, p := range s.peers {
-			if p.Synced() {
-				continue
-			}
 			peers = append(peers, p)
 		}
 		sort.Slice(peers, func(i, j int) bool {
+			if peers[i].Synced() != peers[j].Synced() {
```
ChrisSchinnerl, line 564: @lukechampine this seems to be one possible solution to fix the test but I'd like to hear your input on it first because it doesn't feel like a very clean one.

syncer/syncer_test.go
```diff
@@ -93,3 +93,77 @@ func TestSyncer(t *testing.T) {
...
+			if cm1.Tip() == cm2.Tip() && cm2.Tip() == cm3.Tip() {
+				return
+			}
+			time.Sleep(100 * time.Millisecond)
+		}
+		t.Fatalf("timed out waiting for sync: cm1: %v cm2: %v cm3: %v", cm1.Tip().Height, cm2.Tip().Height, cm3.Tip().Height)
+	}
+	sync()
+
+	testutil.MineBlocks(t, cm2, types.VoidAddress, 20)
```
lukechampine, line 167: this can be fixed by adding:
```go
tipBlock, _ := cm2.Block(cm2.Tip().ID)
s2.BroadcastV2Header(tipBlock.Header())
```

lukechampine:
Leaning towards closing this. Seems like the issue stemmed from a mistaken assumption that the syncer would automatically broadcast mined blocks. We could keep the new test, but I don't think it covers much that isn't already covered by other tests.

n8maninger:
> Leaning towards closing this. Seems like the issue stemmed from a mistaken assumption that the syncer would automatically broadcast mined blocks. We could keep the new test, but I don't think it covers much that isn't already covered by other tests.

I mostly agree. I do think we should consider revisiting the dev experience in the future though. There have been a few foot guns that have bitten us, where the fix has been “call this then this or else.” Maybe just lack of documentation in a couple tricky spots? If it’s biting us, it will definitely bite less experienced developers.

PR #163, opened 2025-01-23
Fix race 

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/coreutils/issues/162

Fixes race in `handleRPCRenewContract` and `handleRPCRefreshContract` which seems to occur due to the ephemeral contractor sharing its internal state without deep-copying.


.changeset/fixed_syncer_deadlocking_when_connect_is_called_after_close.md
ChrisSchinnerl, line 1: Snuck this in here since the merged PR doesn't have one

testutil/host.go
```diff
@@ -75,6 +76,21 @@ func (ec *EphemeralContractor) V2FileContractElement(contractID types.FileContra
...
+	element.EncodeTo(enc)
+	if err := enc.Flush(); err != nil {
+		return types.ChainIndex{}, types.V2FileContractElement{}, err
+	}
+	var copied types.V2FileContractElement
+	dec := types.NewBufDecoder(buf.Bytes())
+	copied.DecodeFrom(dec)
+	if err := dec.Err(); err != nil {
+		return types.ChainIndex{}, types.V2FileContractElement{}, err
+	}
```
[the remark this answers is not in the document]
lukechampine: `element.StateElement.MerkleProof = append([]types.Hash256(nil), element.StateElement.MerkleProof...)` should work.

Perhaps we should add a `StateElement.DeepCopy` method, since this pattern occurs a lot.

testutil/host.go
```diff
@@ -61,11 +61,18 @@ type EphemeralContractor struct {
...
 
-	mu sync.Mutex
+	mu       sync.Mutex
+	shutdown chan struct{}
 }
 
 var _ rhp4.Contractor = (*EphemeralContractor)(nil)
 
+// Close closes the contractor by interrupting its background loop
+func (ec *EphemeralContractor) Close() error {
```
ChrisSchinnerl, line 71: FYI minor change to avoid 200 goroutines in a stacktrace of a CI job running the full test suite

syncer/syncer.go
```diff
@@ -421,19 +425,19 @@ func (s *Syncer) alreadyConnected(id gateway.UniqueID) bool {
 
 func (s *Syncer) acceptLoop(ctx context.Context) error {
 	for {
-		select {
-		case <-ctx.Done():
-			return nil
-		default:
+		ctx, done, err := s.tg.AddContext(ctx)
```
n8maninger, line 428: Better to move this out of the for loop and keep the select.

syncer/syncer.go
```diff
@@ -372,8 +373,11 @@ func (s *Syncer) relayV2TransactionSet(index types.ChainIndex, txns []types.V2Tr
 func (s *Syncer) allowConnect(ctx context.Context, peer string, inbound bool) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if s.closed {
-		return ErrClosed
+
+	select {
```
n8maninger, line 377: Should add to the threadgroup and let it handle it.

threadgroup/threadgroup.go
```diff
@@ -0,0 +1,96 @@
...
+	// wrap the parent context in a cancellable context
+	ctx, cancel := context.WithCancel(parent)
+	// start a goroutine to wait for either the parent context being cancelled
+	// or the threagroup being stopped
+	go func() {
+		select {
+		case <-ctx.Done():
+			break
+		case <-tg.closed:
+			break
```
lukechampine, line 54: breaks are unnecessary here

syncer/syncer.go
```diff
@@ -728,11 +746,7 @@ func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
 		ConnAddr: conn.RemoteAddr().String(),
 		Inbound:  false,
 	}
-	s.wg.Add(1)
-	go func() {
-		defer s.wg.Done()
-		s.runPeer(p)
-	}()
+	go s.runPeer(p)
```
n8maninger, line 749: It seems like `runPeer` also needs a threadgroup at the top.

ChrisSchinnerl: So initially I removed it on purpose cause I thought it's probably a bad idea to prevent `runPeer` from being executed since `Connect` always adds the peer to `s.peers`. Effectively leading to the peer to never getting removed.
But I added it and instead moved the `defer` removal in `runPeer` to the top to make sure it always runs considering that both `Connect` AND `runPeer` add the peer. So it was probably not save to do the `defer` after adding the peer in `runPeer` to begin with and might have been a potential cause for a deadlock on shutdown.

syncer/syncer.go
```diff
@@ -654,17 +671,26 @@ func (s *Syncer) syncLoop(ctx context.Context) error {
 // connections, and syncing the blockchain from active peers. It blocks until an
 // error occurs, upon which all connections are closed and goroutines are
 // terminated.
-func (s *Syncer) Run(ctx context.Context) error {
-	ctx, cancel := context.WithCancel(ctx)
-	s.wg.Add(1)
-	defer s.wg.Done()
+func (s *Syncer) Run() error {
+	ctx, cancel, err := s.tg.AddContext(context.Background())
```
peterjan, line 675: inconsistent with 

```go
        ctx, done, err := s.tg.AddContext(ctx)
	if err != nil {
		return threadgroup.ErrClosed
	}
	defer done()
```

syncer/syncer.go
```diff
@@ -654,17 +671,26 @@ func (s *Syncer) syncLoop(ctx context.Context) error {
...
-	err := <-errChan
-	cancel()
+	for _, fn := range []func(context.Context) error{s.acceptLoop, s.peerLoop, s.syncLoop} {
+		go func() {
+			done, err := s.tg.Add()
+			if err != nil {
+				errChan <- err
+				return
+			}
+			errChan <- fn(ctx)
```
peterjan, line 689: Fine to leave as is but I'd make `errChan` buffered. I had to manually verify whether you follow this up with two `<-errChan`, otherwise we'd potentially be deadlocked again.

syncer/syncer.go
```diff
@@ -691,24 +717,19 @@ func (s *Syncer) Run(ctx context.Context) error {
...
-	s.mu.Lock()
+	s.tg.Stop()
 	return err
 }
 
 // Connect forms an outbound connection to a peer.
 func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
-	if s.closed {
-		return nil, ErrClosed
+	ctx, done, err := s.tg.AddContext(ctx)
```
peterjan, line 727: inconsistent with

```
	ctx, done, err := s.tg.AddContext(ctx)
	if err != nil {
		return threadgroup.ErrClosed
	}
	defer done()
```

peterjan (dismissed):
LGTM 👍

syncer/syncer.go
ChrisSchinnerl: Making it buffered breaks the syncer cause it would not block on the first receive leading to the listener being closed immediately.

peterjan (approved):

PR #166, opened 2025-02-05
mod: Update core dependency

lukechampine:
Replaces `ForEach` calls with `for` loops (see https://github.com/SiaFoundation/core/pull/270)

PR #167, opened 2025-02-05
Fix single address wallet contract payout event IDs

n8maninger:
Follow up from #166 fixing contract payout event IDs


n8maninger:
Closed in favor of #168

PR #173, opened 2025-02-10
wallet: use value receiver on Event.EncodeTo

peterjan:


wallet/events.go
```diff
@@ -340,7 +340,7 @@ func (er *EventV2ContractResolution) DecodeFrom(d *types.Decoder) {
 }
 
 // EncodeTo implements types.EncoderTo
-func (ev *Event) EncodeTo(e *types.Encoder) {
```
n8maninger, line 343: What's the motivation behind this change?

peterjan: I was adding an `encode` function and tried passing an `Event` and noticed it didn't implement `types.EncoderTo`. It's just an inconsistency, obviously I can pass a pointer but figured we might as well change it? In general I feel it's good practice to only use a pointer receiver when you have to, otherwise it usually just triggers me and makes me weary something is going on I'm not aware of.

peterjan:
I'll close the PR but perhaps someone can take it along with them when they make a change to coreutils.

PR #176, opened 2025-02-11
Consider wallet redistributing successful if at least one txn was created

mike76-dev:
I had this issue before with a custom implementation of the single-address wallet. I solved it back then but it seems to have come back when I switched to the `coreutils` implementation.

Both `Redistribute` and `RedistributeV2` iterate over the available inputs when filling up the batch txn; when no more inputs are available, they stop and return an error.

The thing is that, even if not enough outputs were generated by the function, it should still be a success if at least one txn was generated. Because otherwise, the next wallet maintenance will result in the same error and no outputs will be created.

With the fix, the wallet has at least a chance to spawn the required number of outputs within the next few blocks.

### DIFF
--- a/.changeset/consider_wallet_redistributing_successful_if_at_least_one_txn_was_created.md
+++ b/.changeset/consider_wallet_redistributing_successful_if_at_least_one_txn_was_created.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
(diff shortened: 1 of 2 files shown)


n8maninger:
I have no issues with this change

n8maninger (approved):
Might rename `outputs` to `maxOutputs`

PR #180, opened 2025-02-14
Implement RPCReplenishAccounts

n8maninger:
Requires https://github.com/SiaFoundation/core/pull/276

### DIFF
--- a/rhp/v4/rpc_test.go
+++ b/rhp/v4/rpc_test.go
@@ -157,6 +157,26 @@ func mineAndSync(tb testing.TB, cm *chain.Manager, addr types.Address, n int, ti
 	}
 }
 
+func assertValidRevision(t *testing.T, cm rhp4.ChainManager, c rhp4.Contractor, revision rhp4.ContractRevision) {
+	t.Helper()
+
+	basis, fce, err := c.V2FileContractElement(revision.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	revisionTxn := types.V2Transaction{
+		FileContractRevisions: []types.V2FileContractRevision{
+			{
+				Parent:   fce,
+				Revision: revision.Revision,
+			},
+		},
+	}
+	if _, err := cm.AddV2PoolTransactions(basis, []types.V2Transaction{revisionTxn}); err != nil {
+		t.Fatal(err)
+	}
+}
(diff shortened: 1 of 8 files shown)


rhp/v4/server.go
```diff
@@ -436,6 +441,79 @@ func (s *Server) handleRPCFundAccounts(stream net.Conn) error {
...
+	sigHash := s.chain.TipState().ContractSigHash(revision)
+	if !revision.RenterPublicKey.VerifyHash(sigHash, renterSigResp.RenterSignature) {
+		return rhp4.ErrInvalidSignature
+	}
+	revision.RenterSignature = renterSigResp.RenterSignature
+	revision.HostSignature = s.hostKey.SignHash(sigHash)
+
+	_, err = s.contractor.CreditAccountsWithContract(deposits, req.ContractID, revision, usage)
+	if err != nil {
+		return fmt.Errorf("failed to credit account: %w", err)
```
peterjan, line 509:
```suggestion
		return fmt.Errorf("failed to credit accounts: %w", err)
```

peterjan (commented):
LGTM 🔥

rhp/v4/rpc_test.go
```diff
@@ -765,6 +767,184 @@ func TestRPCRenew(t *testing.T) {
...
+		}
+
+		expected, ok := balanceMap[account.Account]
+		if !ok {
+			t.Fatalf("expected account not found")
+		} else if !balance.Equals(expected) {
+			t.Fatalf("expected %v, got %v", expected, balance)
+		}
+	}
+}
```
ChrisSchinnerl, line 946: Considering that we've had this issue in the past for I think renewals can we try and add that new revision to the transaction pool at the end here to make sure it's a valid revision for all parties?

n8maninger: No, I still need to do it. Hold off on re-review please.

n8maninger: I added a helper and the check to everywhere we return a revision. I could move it to a follow-up if you don't like the diff, but it did expose another place we weren't setting the host sig on the renter side.

ChrisSchinnerl (approved):
Great catch in `RPCFreeSectors`

PR #193, opened 2025-03-03
allow v1 basis updates

n8maninger:
We have run into a problem on Zen when the renter and host are not completely synced. There are still miners mining the occasional V1 block causing contract formation to fail. 

@lukechampine since we're storing all the block supplements regardless, is this a safe change to make? After the require height, b.V2 will always be non-nil. Feel free to take over this PR if it is not to your liking.

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -1107,6 +1107,8 @@ func (m *Manager) updateV2TransactionProofs(txns []types.V2Transaction, from, to
 		b, bs, cs, ok := blockAndParent(m.store, index.ID)
 		if !ok {
 			return nil, fmt.Errorf("missing reverted block at index %v", index)
+		} else if bs == nil {
+			bs = new(consensus.V1BlockSupplement)
 		}
 		cru := consensus.RevertBlock(cs, b, *bs)
 		for i := range txns {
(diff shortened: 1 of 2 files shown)


lukechampine (commented):
I *think* this is ok, as long as we are actually storing a block supplement even for V2 blocks. If not, we'd want to do something like:

```go
b, bs, cs, ok := blockAndParent(m.store, index.ID)

if bs == nil {
    bs = new(consensus.V1BlockSupplement)
}
```

ChrisSchinnerl (commented):
Any sort of testing we can perform to go from "I think this is ok" to "This is ok"? 😅 

Doesn't seem like we had a test that ran into the error so at the very least it would be nice to have one that does and after the change doesn't anymore.

lukechampine:
Ok, added a test. It also seems like my suggested safeguards aren't strictly necessary, because we continue constructing supplements even for v2 blocks:
```go
func (db *DBStore) SupplementTipBlock(b types.Block) (bs consensus.V1BlockSupplement) {
	height := db.getHeight()
	if height >= db.n.HardforkV2.RequireHeight {
		return consensus.V1BlockSupplement{Transactions: make([]consensus.V1TransactionSupplement, len(b.Transactions))}
	}
```
still, it doesn't hurt.

ChrisSchinnerl (approved):

peterjan (approved):
LGTM 👍

PR #195, opened 2025-03-04
fix: Add context to siamux Upgrade

ChrisSchinnerl:
Noticed that the `Upgrade` function is still timing out in `renterd` so I checked and noticed that it didn't impose a deadline. 
So to unbreak the interface change I added the `ctx` back and moved the timeout safeguard from `Dial` to `Upgrade`

### DIFF
--- a/rhp/v4/rpc_test.go
+++ b/rhp/v4/rpc_test.go
@@ -856,7 +856,7 @@ func TestSiamuxDialUpgradeTimeout(t *testing.T) {
 		return err != nil && (strings.Contains(err.Error(), "use of closed network connection") || strings.Contains(err.Error(), "operation was canceled"))
 	}
 
-	sr := &blockingSettingsReporter{blockChan: make(chan struct{})}
+	sr := testutil.NewEphemeralSettingsReporter()
 	hk := types.GeneratePrivateKey()
 	rs := rhp4.NewServer(hk, cm, s, c, w, sr, ss, rhp4.WithPriceTableValidity(2*time.Minute))
 	hostAddr := testutil.ServeSiaMux(t, rs, zap.NewNop())
--- a/rhp/v4/rpc_test.go
+++ b/rhp/v4/rpc_test.go
@@ -844,6 +844,48 @@ func TestRPCTimeout(t *testing.T) {
 	})
 }
 
+func TestSiamuxDialUpgradeTimeout(t *testing.T) {
+	n, genesis := testutil.V2Network()
+	cm, s, w := startTestNode(t, n, genesis)
+
+	ss := testutil.NewEphemeralSectorStore()
+	c := testutil.NewEphemeralContractor(cm)
+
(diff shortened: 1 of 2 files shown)


rhp/v4/siamux/siamux.go
```diff
@@ -56,16 +56,20 @@ func Dial(ctx context.Context, addr string, peerKey types.PublicKey) (rhp4.Trans
 	if err != nil {
 		return nil, fmt.Errorf("failed to connect to %q: %w", addr, err)
 	}
-	if deadline, ok := ctx.Deadline(); ok {
-		if err := conn.SetDeadline(deadline); err != nil {
-			return nil, fmt.Errorf("failed to set deadline: %w", err)
-		}
-	}
```
ChrisSchinnerl, line 63: This was actually counter-productive. Setting a deadline on a stream is fine since it serves only a single purpose but the connection potentially lives for a lot longer. So just because we interrupt the dial after 30s, 60s, or even longer, we shouldn't be killing the mux with a deadline after that time.

rhp/v4/rpc_test.go
```diff
@@ -844,6 +844,48 @@ func TestRPCTimeout(t *testing.T) {
...
+
+	ss := testutil.NewEphemeralSectorStore()
+	c := testutil.NewEphemeralContractor(cm)
+
+	isTimeoutErr := func(err error) bool {
+		t.Helper()
+		return err != nil && (strings.Contains(err.Error(), "use of closed network connection") || strings.Contains(err.Error(), "operation was canceled"))
+	}
+
+	sr := &blockingSettingsReporter{blockChan: make(chan struct{})}
```
peterjan, line 859: unused now right?

peterjan:
looks good but we need the same thing for QUIC right?

peterjan (approved):

rhp/v4/siamux/siamux.go
```diff
@@ -56,16 +56,20 @@ func Dial(ctx context.Context, addr string, peerKey types.PublicKey) (rhp4.Trans
...
-	return Upgrade(conn, peerKey)
+	return Upgrade(ctx, conn, peerKey)
 }
 
 // Upgrade upgrades an existing connection to use the SiaMux transport.
-func Upgrade(conn net.Conn, peerKey types.PublicKey) (rhp4.TransportClient, error) {
+func Upgrade(ctx context.Context, conn net.Conn, peerKey types.PublicKey) (rhp4.TransportClient, error) {
```
n8maninger, line 63: If the context doesn't have a deadline, we should still set a deadline on the conn, similar to what we did with the stream. Probably only like 5-10s. You can clear it after the upgrade is complete with `conn.SetDeadline(time.Time{})`.

n8maninger (changes requested):
Should add a deadline before `mux.Accept` and clear it after the handshake is complete in `Serve` too.

peterjan (approved):

PR #199, opened 2025-03-13
Add default dial timeout

n8maninger:
Adds a default dial timeout for the SiaMux transport

### DIFF
--- a/.changeset/added_a_default_deadline_of_1m_to_the_siamux_dialer_if_the_context_has_a_deadline_it_will_override_the_default.md
+++ b/.changeset/added_a_default_deadline_of_1m_to_the_siamux_dialer_if_the_context_has_a_deadline_it_will_override_the_default.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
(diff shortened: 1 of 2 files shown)


lukechampine:
For me, the implicit contract with a context is, "if I don't cancel you, the operation won't be canceled." It would be kinda surprising for me if I passed in a context that only had a `WithCancel`, and the function failed with `deadline exceeded`.

I can see how this would be convenient for code that's receiving a context from the caller and passing it through to `Dial`; now they don't have to add a `WithDeadline` to it. But idk, that feels like a reasonable ask to me.

n8maninger:
I can understand that. To me, a context on net io gives the caller an option to "own" the deadline over defaulting to no deadline since there's really no good reason to block forever.

ChrisSchinnerl (commented):
Tbh I feel pretty indifferent about this change. If we didn't do the whole checking a context for a deadline thing I'd say let's not add a timeout here and rely only on contexts. But since we already do both we might as well add a separate timeout to the dial for completion.

peterjan (dismissed):
I'll approve it!

ChrisSchinnerl (approved):

peterjan (approved):

PR #201, opened 2025-03-17
chore(deps): add erravimus testnet consts

n8maninger:


chain/network.go
```diff
@@ -186,3 +186,27 @@ func TestnetAnagami() (*consensus.Network, types.Block) {
 	}}
 	return n, genesis
 }
+
+func TestnetEravimus() (*consensus.Network, types.Block) {
```
peterjan, line 190: typo `TestnetErravimus`

ChrisSchinnerl: indeed

ChrisSchinnerl (changes requested):

PR #204, opened 2025-03-26
DB migration

lukechampine:
This adds a generic `MigrateDB` function for updating consensus databases, and a v1->v2 migration that recomputes the full element tree. I haven't run a full mainnet migration yet, but migrating the first 200k blocks took ~12 minutes, so likely around 30 minutes for the entire chain.

### DIFF
--- a/.changeset/db_migration.md
+++ b/.changeset/db_migration.md
@@ -1,0 +1,7 @@
+---
+default: CHANGE_TYPE
+---
+
+# DB migration
+
+This adds a generic `MigrateDB` function for updating consensus databases, and a v1->v2 migration that recomputes the full element tree, which fixes Zen.
--- a/.changeset/db_migration.md
+++ b/.changeset/db_migration.md
@@ -1,5 +1,5 @@
 ---
-default: CHANGE_TYPE
+default: major
 ---
 
 # DB migration
--- a/chain/db.go
+++ b/chain/db.go
@@ -757,8 +757,9 @@ func (db *DBStore) Flush() error {
 }
 
 // NewDBStore creates a new DBStore using the provided database. The tip state
-// is also returned. The DB will be automatically migrated if necessary.
-func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBStore, _ consensus.State, err error) {
+// is also returned. The DB will be automatically migrated if necessary. The
+// provided logger may be nil.
+func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block, logger MigrationLogger) (_ *DBStore, _ consensus.State, err error) {
 	// during initialization, we should return an error instead of panicking
 	defer func() {
 		if r := recover(); r != nil {
@@ -807,7 +808,10 @@ func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBSto
 			return nil, consensus.State{}, err
 		}
 	} else if version[0] != 2 {
-		if err := MigrateDB(db, n, noopLogger{}); err != nil {
+		if logger == nil {
+			logger = noopLogger{}
+		}
+		if err := migrateDB(dbs, n, logger); err != nil {
 			return nil, consensus.State{}, fmt.Errorf("failed to migrate database: %w", err)
 		}
 	}
--- a/chain/db.go
+++ b/chain/db.go
@@ -474,7 +474,7 @@ func (db *DBStore) revertState(prev consensus.State) {
 
 func (db *DBStore) applyElements(cau consensus.ApplyUpdate) {
 	cau.ForEachTreeNode(func(row, col uint64, h types.Hash256) {
-		db.bucket(bTree).put(db.treeKey(row, col), h)
+		db.bucket(bTree).putRaw(db.treeKey(row, col), h[:])
 	})
 
 	for _, sced := range cau.SiacoinElementDiffs() {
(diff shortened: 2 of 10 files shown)


chain/migrate.go
```diff
@@ -0,0 +1,97 @@
...
+			tree = tree[:0]
+			return db.Flush()
+		}
+		lastPrint := time.Now()
+		for height := range v1Blocks {
+			index, ok0 := dbs.BestIndex(height)
+			_, b, bs, ok1 := dbs.getBlock(index.ID)
+			ancestorTimestamp, ok2 := dbs.AncestorTimestamp(b.ParentID)
+			if !ok0 || !ok1 || !ok2 {
+				return errors.New("database is corrupt")
```
peterjan, line 60: context could be useful in the unlikely event someone runs into this

chain/migrate.go
```diff
@@ -0,0 +1,97 @@
...
+		v1Blocks := min(dbs.getHeight(), n.HardforkV2.RequireHeight)
+		seen := make(map[[2]uint64]types.Hash256)
+		var tree [][2][]byte
+		flush := func() error {
+			sort.Slice(tree, func(i, j int) bool {
+				return bytes.Compare(tree[i][0], tree[j][0]) < 0
+			})
+			bucket := dbs.bucket(bTree)
+			for _, kv := range tree {
+				bucket.putRaw(kv[0], kv[1])
```
peterjan, line 48: Killing the migration mid-process is a non issue right?

peterjan (commented):
looks good to me 🔥

chain/db.go
```diff
@@ -798,7 +797,12 @@ func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBSto
 		if err := dbs.Flush(); err != nil {
 			return nil, consensus.State{}, err
 		}
-	} else if dbGenesis.ID != genesisBlock.ID() {
+	} else if version[0] != 2 {
+		return nil, consensus.State{}, errors.New("incompatible version; please migrate the database")
```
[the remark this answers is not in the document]
lukechampine: Seemed a bit rude to possibly kick off an arbitrarily-long migration from a constructor function. (Plus you'd have to pass in a `MigrationLogger`.) I agree that it's not an ideal API, though. What do you do for DB migrations in renterd?

chain/migrate.go
lukechampine: in this case, yeah; if you fail in the middle, you lose all progress and start over from 0. If we add a much more expensive migration in the future, we'd probably want to track its progress in a temporary DB bucket or something.

chain/db.go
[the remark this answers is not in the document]
ChrisSchinnerl: Our nodes all migrate implicitly because we assume that people run automated setups. So we don't want to require user interaction at all. To be fair most migrations are also quite fast. As the network matures we might need to consider zero downtime migrations.

That said I think returning an error might be less assuming on our end but also leads to downtime. At the end of the day we definitely don't want people to hold off on that migration ^^

lukechampine:
@n8maninger thoughts? I'm sympathetic to Chris but it would be kinda annoying to change the function signature (which we could avoid if we don't pass a `MigrationLogger`, but then you have no feedback for long migrations).

Wait, here's an idea: have `NewDBStore` call `MigrateDB` with a no-op logger. That way, we get migrate-by-default, and if you want feedback, you can call `MigrateDB` explicitly on startup.

lukechampine:
it's a bit different from a typical "logger" because I wanted to include the notion of a progress bar, e.g. to feed into a UI. Maybe there's a way to do that with zap, though.

lukechampine:
ok, implemented automatic migration 👍🏻

chain/db.go
```diff
@@ -807,7 +806,14 @@ func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBSto
 		if err := dbs.Flush(); err != nil {
 			return nil, consensus.State{}, err
 		}
-	} else if dbGenesis.ID != genesisBlock.ID() {
+	} else if version[0] != 2 {
+		if err := MigrateDB(db, n, noopLogger{}); err != nil {
```
peterjan, line 810: Isn't it worth passing in a logger to `NewDBStore` to make us aware a migration is taking place? Definitely a little annoying but the way it is now this would hang for a really long time without any visible trace of what's going on. 
User might decide to cancel out and would have to start from scratch every time.

ChrisSchinnerl: +1 I'd consider Nate's suggested approach of a functional argument for a zap logger

lukechampine: If we use a functional argument, then the API won't break, but you still need to update all your callsites to pass a logger. I don't see how that's much better than updating all your callsites to call `MigrateDB`.
I'm not really a fan of functional arguments in general; they remove the "cost" of adding new parameters to a function, without which things tend to complexify without bound.

Anyway, the important thing is to get this merged so that Zen can stop being broken, so I'll do whatever you guys feel is most expedient.

ChrisSchinnerl: I don't mind breaking the signature but I do vote for the migration being implicit and probably for the no-op logger to be exposed. It's quite annoying to have a dependency break and to find out you need to implement some interface to satisfy a function call. Might as well make it painless for people who don't care about logging.

Regarding complexity I'd argue that figuring out a function argument to enable logging has lower complexity than figuring out you need to call a separate method to do the same 😅

.changeset/db_migration.md
```diff
@@ -0,0 +1,7 @@
+---
+default: CHANGE_TYPE
```
ChrisSchinnerl, line 2: Hey @<EMPLOYEE_NAME> I think you missed something here 😅

PR #207, opened 2025-04-04
Return up to `maxBlocks` updates, not up to `maxBlocks+1`

mike76-dev:
This is not a big deal, but if you request like 100 updates per batch, you would normally expect the last synced index to grow like `100, 200, 300, ...` and not like `101, 202, 303, ...`.


lukechampine (approved):

lukechampine:
good catch!

PR #210, opened 2025-04-12
More aggressive migration

lukechampine:
Previous piecemeal attempts at migration produced further discrepancies in the DB. Here we avoid this by taking a scorched-earth approach, nuking basically everything in the DB and resyncing from ourselves. Since sidechains cannot easily be fixed the same way, they are simply deleted. This produces a final state equivalent to resyncing from a good peer.

I have tested this with both Zen and Erravimus consensus.dbs, and it appears to successfully repair them to a valid state. (I reached block 113487 on Zen, and 3311 on Erravimus.) I suggest we deploy this to Erravimus, confirm that we can continue mining there, then deploy to Zen miners. Note that existing Zen blocks above 113487 will (eventually) be reorged.

Performance-wise, there is room for improvement here. The old migration code achieved significant speedups by caching DB writes in memory and sorting them before flushing. I am tempted to apply a similar strategy to the `DBStore` itself, in a more generic fashion -- but this is slightly risky and not critical, so if I do explore it, it'll be on a different branch.

### DIFF
--- a/chain/db.go
+++ b/chain/db.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"iter"
 	"math/bits"
 	"time"
 
@@ -70,6 +71,7 @@ type DBBucket interface {
 	Get(key []byte) []byte
 	Put(key, value []byte) error
 	Delete(key []byte) error
+	Iter() iter.Seq2[[]byte, []byte]
 }
 
 // MemDB implements DB with an in-memory map.
@@ -171,6 +173,20 @@ type memBucket struct {
 func (b memBucket) Get(key []byte) []byte       { return b.db.get(b.name, key) }
 func (b memBucket) Put(key, value []byte) error { return b.db.put(b.name, key, value) }
 func (b memBucket) Delete(key []byte) error     { return b.db.delete(b.name, key) }
+func (b memBucket) Iter() iter.Seq2[[]byte, []byte] {
+	return func(yield func([]byte, []byte) bool) {
+		for key, val := range b.db.buckets[b.name] {
+			if pval, ok := b.db.puts[b.name][string(key)]; ok {
+				val = pval
+			} else if _, ok := b.db.dels[b.name][string(key)]; ok {
+				continue
+			}
+			if !yield([]byte(key), val) {
+				return
+			}
+		}
+	}
+}
 
 // NewMemDB returns an in-memory DB for use with DBStore.
 func NewMemDB() *MemDB {
@@ -614,7 +630,7 @@ func (db *DBStore) SupplementTipTransaction(txn types.Transaction) (ts consensus
 		if fce, ok := db.getFileContractElement(sp.ParentID, numLeaves); ok {
 			if windowIndex, ok := db.BestIndex(fce.FileContract.WindowStart - 1); ok {
 				ts.StorageProofs = append(ts.StorageProofs, consensus.V1StorageProofSupplement{
-					FileContract: fce.Copy(),
+					FileContract: fce.Move(),
 					WindowID:     windowIndex.ID,
 				})
 			}
@@ -656,8 +672,6 @@ func (db *DBStore) SupplementTipBlock(b types.Block) (bs consensus.V1BlockSupple
 func (db *DBStore) AncestorTimestamp(id types.BlockID) (t time.Time, ok bool) {
 	cs, _ := db.State(id)
 	if cs.Index.Height > db.n.HardforkOak.Height {
-		// the Oak difficulty adjustment algorithm is continuous, so ancestor
-		// timestamps are not needed
 		return time.Time{}, true
 	}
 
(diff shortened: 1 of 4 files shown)


ChrisSchinnerl (commented):
Code looks good but crashed my Zen node (16GB RAM) right before finishing. Seems to work now but requiring a restart is not the best UX. Trying to run this on my mainnet node now.

chain/migrate.go
```diff
@@ -50,50 +48,66 @@ func NewZapMigrationLogger(log *zap.Logger) MigrationLogger {
...
 		}
+		for id := range toDelete {
+			dbs.bucket(bBlocks).delete(id[:])
+			dbs.bucket(bStates).delete(id[:])
+		}
+		l.Printf("Removing block supplement data")
+		for id := range dbs.db.Bucket(bFileContractElements).Iter() {
+			dbs.bucket(bFileContractElements).delete(id)
+		}
+		for id := range dbs.db.Bucket(bSiacoinElements).Iter() {
```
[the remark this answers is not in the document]
lukechampine: In practice it seems relatively quick even for a large bucket, and I'd rather not add another method.

lukechampine:
> Code looks good but crashed my Zen node (16GB RAM) right before finishing
Hmm. I wonder if it was the "removing subsequent blocks" code that crashed it? I assumed that it was okay to delete a bunch of keyvals without flushing, but maybe not.

chain/migrate.go
```diff
@@ -50,50 +48,66 @@ func NewZapMigrationLogger(log *zap.Logger) MigrationLogger {
...
+		}
+		l.Printf("Removing tree data")
+		for k := range dbs.db.Bucket(bTree).Iter() {
+			dbs.bucket(bTree).delete(k)
+		}
+
+		l.Printf("Recomputing main chain")
+		v1Blocks := min(dbs.getHeight(), n.HardforkV2.RequireHeight) + 1
+		cs := n.GenesisState()
 		lastPrint := time.Now()
```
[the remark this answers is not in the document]
lukechampine: fair enough

peterjan (dismissed):

chain/migrate.go
```diff
@@ -43,74 +41,109 @@ func (zl *zapMigrationLogger) SetProgress(percentage float64) {
...
+		for id := range dbs.db.Bucket(bSiafundElements).Iter() {
+			dbs.bucket(bSiafundElements).delete(id)
+			if err := flushDeletes(); err != nil {
+				return err
+			}
+		}
+		l.Printf("Removing tree data")
+		for k := range dbs.db.Bucket(bTree).Iter() {
+			dbs.bucket(bTree).delete(k)
+			if dels++; dels%10000 == 0 {
```
peterjan, line 97: use `flushDeletes`

chain/db.go
```diff
@@ -181,6 +198,135 @@ func NewMemDB() *MemDB {
...
+	err := b.mb.Put(key, value)
+	return err
+}
+
+func (b cacheBucket) Delete(key []byte) error {
+	return b.mb.Delete(key)
+}
+
+func (b cacheBucket) Iter() iter.Seq2[[]byte, []byte] {
+	return b.db.Iter() // NOTE: imperfect
```
peterjan, line 225: 😬, maybe add disclaimer to `NewCacheDB`?

chain/db.go
```diff
@@ -181,6 +198,135 @@ func NewMemDB() *MemDB {
...
+		}
+		db.kvs[name] = bucket
+	}
+	for bucket, kvs := range db.kvs {
+		bucket := db.db.Bucket([]byte(bucket))
+		sort.Slice(kvs, func(i, j int) bool {
+			return bytes.Compare(kvs[i][0], kvs[j][0]) < 0
+		})
+		for _, kv := range kvs {
+			bucket.Put(kv[0], kv[1])
```
peterjan, line 271: Might be overlooking something but is there a reason we don't process `dels` to potentially save a db write here?

chain/db.go
```diff
@@ -715,8 +860,8 @@ func (db *DBStore) PruneBlock(id types.BlockID) {
 func (db *DBStore) shouldFlush() bool {
 	// NOTE: these values were chosen empirically and should constitute a
 	// sensible default; if necessary, we can make them configurable
-	const flushSizeThreshold = 20e6
-	const flushDurationThreshold = 1000 * time.Millisecond
+	const flushSizeThreshold = 100e6
```
peterjan, line 863: `nit` Feels a bit high? For deletes we only count the key length so if the value is 1kb each this is roughly 3GB.

peterjan: right! woopsie

chain/db.go
```diff
@@ -181,6 +198,135 @@ func NewMemDB() *MemDB {
...
+		for _, kv := range kvs {
+			bucket.Delete(kv[0])
+		}
+	}
+	for name := range db.kvs {
+		clear(db.kvs[name])
+		db.kvs[name] = db.kvs[name][:0]
+	}
+
+	// clear MemDB
```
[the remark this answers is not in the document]
lukechampine: yeah, I thought about it. Easy to do later if we end up needing it in more than one place.

chain/db.go
lukechampine: I'll just implement it properly lol

PR #215, opened 2025-04-24
Pass MissedHostValue to Validate

ChrisSchinnerl:
I think we made a mistake here. `Validate` expects the existing unallocated collateral and the unallocated collateral should be the `MissedHostValue` rather than `TotalCollateral-MissedHostValue`. After all, when we pay with a contract, that's the field we subtract from.

### DIFF
--- a/rhp/v4/rpc_test.go
+++ b/rhp/v4/rpc_test.go
@@ -543,14 +543,31 @@ func TestRPCRefresh(t *testing.T) {
 
 	t.Run("valid refresh", func(t *testing.T) {
 		revision := formContractUploadSector(t, types.Siacoins(100), types.Siacoins(200), types.Siacoins(25))
+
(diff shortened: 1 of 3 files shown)


rhp/v4/rpc_test.go
```diff
@@ -543,16 +543,33 @@ func TestRPCRefresh(t *testing.T) {
 
 	t.Run("valid refresh", func(t *testing.T) {
 		revision := formContractUploadSector(t, types.Siacoins(100), types.Siacoins(200), types.Siacoins(25))
+
+		// compute the min allowance the host will accept
```
n8maninger, line 547: There shouldn't be a need to change this test right? Forming a contract with 100/200 SC then refreshing it to add 10/20 should be a totally valid ratio.

PR #216, opened 2025-04-24
Add error to relay transaction set

n8maninger:
This adds an error to the syncer methods and changes them to block until at least one relay returns successful. This won't guarantee that the transaction set is actually accepted by any peers, but it makes it more likely that at least one other peer received it.

Also noticed that we ban peers that relay transaction sets with no transactions, while it's a fair practice I added a check to prevent "good" clients from being banned since we rarely check length before calling broadcast.

This may help with debugging random broadcast issues (https://github.com/SiaFoundation/hostd/issues/646)

### DIFF
--- a/.changeset/return_error_when_transaction_set_relay_fails.md
+++ b/.changeset/return_error_when_transaction_set_relay_fails.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
+# Return error when transaction set relay fails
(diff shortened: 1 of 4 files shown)


n8maninger:
> > Also noticed that we ban peers that relay transaction sets with no transactions, while it's a fair practice I added a check to prevent "good" clients from being banned since we rarely check length before calling broadcast.
> 
> Not in the diff?

https://github.com/SiaFoundation/coreutils/pull/216/files#diff-e725a03bd475dc4ea28028613ede1fd88a8cc8f881f0a1788a189e766f191830R348-R350

lukechampine:
Notably this changes `Broadcast` to be a blocking call rather than async. That's fine, but something to be aware of; there may be places where we want to call `Broadcast` in a goroutine now.

lukechampine (approved):

PR #222, opened 2025-04-29
debug deadlock

peterjan:


syncer/syncer.go
```diff
@@ -813,21 +826,10 @@ func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
...
-			return
-		}
-		s.runPeer(p)
-		done()
-	}()
+	if _, err := s.runPeer(p); err != nil {
+		return nil, fmt.Errorf("failed to add peer: %w", err)
+	}
 
-	// runPeer does this too, but doing it outside the goroutine prevents a race
```
peterjan, line 825: My PR re-introduces this race. I wanted to draw attention to this, I did that because I don't see why it's that much of an issue. But also because fixing that would require `runPeer` to signal the peer was added, as well as be able to execute it in a blocking fashion. Which we could do return two channels but since it's already a bit of anti-pattern I didn't dare pushing that into review 😅

ChrisSchinnerl: What I don't get is why introduce this anti-pattern instead of breaking up `runPeer` into 2 methods to avoid the channel?

1. `addPeer` -> calls `pm.AddPeer`, `pm.UpdatePeerInfo` and adds the peer to `s.peers` so basically all the things we want to wait for
2. `runPeer` -> spawns the goroutine to handle the peer and handle the asynchronous stuff

Then you can call `addPeer` and check for an error indicating whether the peer was added or not before launching `runPeer` in a goroutine. So you don't need to add the peer here anymore.

I think than you can also keep most of the remaining codebase untouched assuming these lines were the culprit causing the issue.

peterjan: Well because I personally prefer the channel to be able to span the goroutine that handles the removal as close as possible to where we add the peer to `s.peers` 🤷🏻‍♂️ . That was before I made this comment though... I'll update the PR

syncer/syncer.go
```diff
@@ -818,16 +823,14 @@ func (s *Syncer) Connect(ctx context.Context, addr string) (*Peer, error) {
 		if err != nil {
 			return
 		}
+		defer done()
+
+		if err := s.addPeer(p); err != nil {
```
ChrisSchinnerl, line 828: This should happen outside the goroutine now. Otherwise you still get the race.

You'll also need to move `s.tg.Add` into `runPeer`, directly after the `defer` delete of the peer. To make sure we always `runPeer` when `addPeer` returns `nil`.

ChrisSchinnerl (changes requested):

syncer/syncer.go
peterjan: Too hasty 👍 thanks

ChrisSchinnerl (approved):
Looks ok to me. Did you confirm that the latest commit still fixes the issue?

peterjan:
> > Looks ok to me. Did you confirm that the latest commit still fixes the issue?

To be very clear, I can not reproduce the deadlock in `renterd` because it happens once in a blue moon and we have the occasional NDF in our integration tests that prevent it from happening. We can safely merge this, it's clearly a potential cause for a deadlock and therefor the likely culprit.

n8maninger (approved):

PR #226, opened 2025-05-12
Persist utxos in the wallet store

### LINKED ISSUE #35: Persist locked outputs in the wallet
The outputs in `w.locked` map are not persisted. If the wallet is shut down after a transaction is broadcast and then restarted immediately (before the next block is mined), the wallet may try to spend an already spent output, because it's unaware.

n8maninger:
Closes #35

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -62,6 +62,15 @@ type (
 		// WalletEventCount returns the total number of events relevant to the
 		// wallet.
 		WalletEventCount() (uint64, error)
+
+		// LockUTXOs locks the given siacoin output IDs for use in a transaction.
+		// The outputs will be unlocked after the given time.
+		LockUTXOs([]types.SiacoinOutputID, time.Time) error
+		// LockedUTXOs returns a list of locked siacoin output IDs. That are locked
+		// as of the given time.
+		LockedUTXOs(time.Time) ([]types.SiacoinOutputID, error)
+		// ReleaseUTXOs unlocks the given siacoin output IDs.
+		ReleaseUTXOs([]types.SiacoinOutputID) error
 	}
 
 	// A SingleAddressWallet is a hot wallet that manages the outputs controlled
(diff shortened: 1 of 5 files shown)


wallet/wallet.go
```diff
@@ -834,30 +873,42 @@ func (sw *SingleAddressWallet) RedistributeV2(outputs int, amount, feePerByte ty
...
 		}
 	}
 	for _, txn := range v2txns {
 		for _, in := range txn.SiacoinInputs {
 			delete(sw.locked, in.Parent.ID)
+			released = append(released, in.Parent.ID)
 		}
 	}
+
+	if err := sw.store.ReleaseUTXOs(released); err != nil {
```
peterjan, line 908: I would collect the ids, update database and then delete them from memory. If `store.ReleaseUTXOs` errors out you'll have deleted them from `sw.locked` already.

ChrisSchinnerl: While I would generally agree that in most cases you'd do I/O first and then the in-memory update for consistency, it might be detrimental here.

If `ReleaseInputs` fails, nobody is going to retry it similar to how nobody retries closing a file if it returns an error. But we know that people still want to be able to spend that input again. So to avoid locking that input up I think releasing from the map might still make sense here.

Of course whether the error is temporary or not is a whole different question and depends on the implementation.

wallet/wallet_test.go
```diff
@@ -363,6 +364,68 @@ func TestWallet(t *testing.T) {
...
+	} else if len(locked) != len(toSign) {
+		t.Fatalf("expected %v locked outputs, got %v", len(toSign), len(locked))
+	}
+
+	for _, id := range toSign {
+		if !slices.Contains(locked, types.SiacoinOutputID(id)) {
+			t.Fatalf("expected locked output %v, got %v", id, locked)
+		}
+	}
+
```
ChrisSchinnerl, line 416: can you extend this test here to create a new wallet from the same store to make sure it's loading the locked inputs from the store upon creation?

ChrisSchinnerl (commented):
Did you forget to push your changes before re-requesting reviews?

peterjan (approved):

ChrisSchinnerl (approved):

PR #234, opened 2025-05-16
Consistently return `rhp4.ErrInvalidSignature`

peterjan:
If we want to store revisions in the database to avoid fetching the latest revision right before revising the contract, we need the host to return `rhp4.ErrInvalidSignature` consistently so we can check for it. Our RPCs expect two types of signatures, a challenge signature that proves we can modify the contract, or a signature that covers the new revision. This PR updates 3 locations where we weren't returning `rhp4.ErrInvalidSignature` when we failed to validate the challenge signature.

Related to https://github.com/SiaFoundation/indexd/issues/144

### DIFF
--- a/.changeset/return_rhp4errinvalidsignature_consistently_when_we_fail_to_validate_the_challenge_signature.md
+++ b/.changeset/return_rhp4errinvalidsignature_consistently_when_we_fail_to_validate_the_challenge_signature.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (commented):
🧪

peterjan:
> 🧪

covered
- `RPCAppendSectors`
- `RPCFundAccounts`
- `RPCReplenishAccounts`
- `RPCFreeSectors`
- `RPCSectorRoots`

PR #235, opened 2025-05-19
Consistently use ContractSigner interface

n8maninger:
Changes RPCAppendSectors and RPCFreeSectors to use `ContractSigner` like the other methods that require a contract signature


peterjan (approved):
👍  nice

ChrisSchinnerl (approved):

PR #244, opened 2025-05-28
Fix race when returning tip

peterjan:


wallet/wallet.go
```diff
@@ -1011,14 +1011,14 @@ func NewSingleAddressWallet(priv types.PrivateKey, cm ChainManager, store Single
 
 	sw := &SingleAddressWallet{
 		priv: priv,
+		addr: types.StandardUnlockHash(priv.PublicKey()),
```
ChrisSchinnerl, line 1014: Unrelated refactor

ChrisSchinnerl (changes requested):

wallet/wallet.go
peterjan: 🙄

ChrisSchinnerl (dismissed):

lukechampine (approved):

PR #245, opened 2025-05-28
Use separate mutex for locked outputs

peterjan:
Cover the locked outputs by its own mutex to avoid lock contention. There were two cases where we needlessly acquire the mutex lock. Having a separate mutex the locking more clear overall. This fixes a deadlock in `renterd`, I tested it manually by running this commit and the test suite was immediately green with this fix.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -85,12 +85,14 @@ type (
 
 		cfg config
 
-		mu  sync.Mutex // protects the following fields
-		tip types.ChainIndex
 		// locked is a set of siacoin output IDs locked by FundTransaction. They
 		// will be released either by calling Release for unused transactions or
 		// being confirmed in a block.
-		locked map[types.SiacoinOutputID]time.Time
+		lockedMu sync.Mutex
+		locked   map[types.SiacoinOutputID]time.Time
+
+		mu  sync.Mutex
+		tip types.ChainIndex
 	}
 )
 
@@ -166,8 +168,8 @@ func (sw *SingleAddressWallet) Balance() (balance Balance, err error) {
 		}
 	}
 
-	sw.mu.Lock()
-	defer sw.mu.Unlock()
+	sw.lockedMu.Lock()
+	defer sw.lockedMu.Unlock()
 	bh := sw.cm.TipState().Index.Height
 	for _, sco := range outputs {
 		if sco.MaturityHeight > bh {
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -488,7 +488,7 @@ func (sw *SingleAddressWallet) FundV2Transaction(txn *types.V2Transaction, amoun
 	if err := sw.lockUTXOs(toLock); err != nil {
 		return types.ChainIndex{}, nil, fmt.Errorf("failed to lock UTXOs: %w", err)
 	}
-	return sw.tip, toSign, nil
+	return tip, toSign, nil
```
n8maninger, line 491: Grabbing tip under lock is intentional so that the proof basis in the database doesn't change out from under us.

ChrisSchinnerl: Don't we already have a bit of an issue with `UnspentSiacoinElements` not being called under the lock? We lock the outputs under the lock but we don't fetch them under it. Might also be true for v1 and other places where we call `UnspentSiacoinElements`.

.changeset/improve_locking_behaviour_by_removing_the_tip_from_memory_and_returning_it_alongside_the_unspent_elements_from_the_store.md
```diff
@@ -0,0 +1,5 @@
+---
+default: minor
```
ChrisSchinnerl, line 2: (nit) strictly speaking it's a breaking change which makes it major. Which ends up being a minor version bump in knope anyway though cause we haven't reached v1.0.0 yet.

ChrisSchinnerl (approved):

.changeset/improve_locking_behaviour_by_removing_the_tip_from_memory_and_returning_it_alongside_the_unspent_elements_from_the_store.md
n8maninger: Lets make it breaking and improve it with a subsection on the changes that were made. i.e.

```
# Improve locking behaviour by removing the tip from memory and returning it alongside the unspent elements from the store.

Changes the store interface for the `SingleAddressWallet` to return the tip. This must be atomic with the current state of the proofs stored in the database.

File: wallet/wallet.go [hunk 3]
```diff
@@ -380,14 +371,14 @@ func (sw *SingleAddressWallet) FundTransaction(txn *types.Transaction, amount ty
 		return nil, nil
 	}
 
-	elements, err := sw.store.UnspentSiacoinElements()
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	elements, _, err := sw.store.UnspentSiacoinElements()
```
n8maninger, line 377: There's no reason to pass the elements into `selectUTXOs` over calling it within the function now.

PR #249, opened 2025-06-02
syncer: Add syncing fallback

lukechampine:
It seems that syncing sometimes stalls even when we have peers with a longer chain. We should investigate exactly why this happens, but in the meantime, add a fallback that resets the peer's sync status after 1-2 minutes, causing us to try to sync with them again later on.


peterjan (approved):
still syncing on my machine, but looking good so far

PR #254, opened 2025-06-07
Fix panic when updating proofs

n8maninger:
Don't panic when attempting to update proofs and return an error instead.


n8maninger:
@lukechampine @ChrisSchinnerl @peterjan merging this to put out a hot fix, but please still review

Issue #256, opened 2025-06-08
Panic in Manager.UnconfirmedParents


mike76-dev:
`UnconfirmedParents` panicking when trying to form a V1 contract:
```
panic: runtime error: index out of range [11] with length 5
goroutine 9302 [running]:
go.sia.tech/coreutils/chain.(*Manager).UnconfirmedParents.func1({0xbe, 0x11, 0x1f, 0x6a, 0xe6, 0x30, 0x10, 0xb7, 0x4b, 0xfd, ...})
         .../go/pkg/mod/go.sia.tech/coreutils@v0.16.1-0.20250607184904-8c78c1233ed4/chain/manager.go:955 +0x1c5
go.sia.tech/coreutils/chain.(*Manager).UnconfirmedParents.func2({{0xc00a11c000, 0x1, 0x1}, {0xc00476e2d0, 0x1, 0x1}, {0xc00477a000, 0x1, 0x1}, {0x0, ...}, ...})
         .../go/pkg/mod/go.sia.tech/coreutils@v0.16.1-0.20250607184904-8c78c1233ed4/chain/manager.go:960 +0x8c
go.sia.tech/coreutils/chain.(*Manager).UnconfirmedParents(_, {{0xc00a11c000, 0x1, 0x1}, {0xc00476e2d0, 0x1, 0x1}, {0xc00477a000, 0x1, 0x1}, ...})
         .../go/pkg/mod/go.sia.tech/coreutils@v0.16.1-0.20250607184904-8c78c1233ed4/chain/manager.go:974 +0x1fd
```
This happens when I'm running this code:
```
return append(cm.UnconfirmedParents(txn), txn), nil
```

lukechampine:
My gut says this is due to the cached `parentMap` not being invalidated when it should be. Perhaps we should be setting `parentMap = nil` in `revalidatePool`?

n8maninger:
Fixed by #260

PR #257, opened 2025-06-09
Remove parentMap

ChrisSchinnerl:
I'm proposing to remove the `parentMap` cache. I think Luke is right in the linked issue that the panic stems from an inconsistency between the cache and pool state. However, if we invalidate the cache every time we call `revalidatePool`, which we call every time we call `UnconfirmedParents`, there is no point in keeping the map anymore. Since it will be invalidated before every time we use it.

Fixes https://github.com/SiaFoundation/coreutils/issues/256

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -99,7 +99,6 @@ type Manager struct {
 		ms             *consensus.MidState
 		weight         uint64
 		medianFee      *types.Currency
-		parentMap      map[types.Hash256]int
 		lastReverted   []types.Transaction
 		lastRevertedV2 []types.V2Transaction
 	}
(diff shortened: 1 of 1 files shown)


lukechampine:
tbc, I meant we should invalidate the cache *if the pool is changed* by `revalidatePool`. If you're calling `UnconfirmedParents` many times between txpool changes, the cache is still helpful.

That said, I never actually benchmarked how long it takes to rebuild the `parentMap`. If the answer is <100ms, then yeah I agree for simplicity we should just remove it.

lukechampine:
Ok, I benchmarked it: with 1000 txns in the pool, each spending 1 input and creating 1 output, it takes <1 ms. With 100 txns, each spending 1 input and creating 100 outputs, it takes 87ms. So things start to get worryingly slow around 10,000 total outputs across txns in the pool.

However, that's for v1 transactions, where computing an output ID requires hashing the full transaction. When I rerun the benchmark with v2 transactions, I see a ~20x speedup: 4ms for 10k outputs. (And about the same <1ms for 1000 single-output txns, as expected.) Given that most of the network has already transitioned to using only v2 transactions, I think we can safely say that we have enough headroom to ditch the map. The only reason to keep it around would be if someone out there needs to call `V2TransactionSet` in a hot loop, which seems unlikely.

lukechampine:
Closing in favor of #260

PR #260, opened 2025-06-09
TxPool fixes

lukechampine:
In the course of writing a test for full txpool behavior, I found a bug in the index handling: all of the transactions were being assigned an index of `len(txpool.txns)`, instead of sequential ones! This was almost certainly the result of copy-pasting the `m.txpool.indices[txid] = len(m.txpool.txns)` from `AddPoolTransactions`; it breaks in `revalidatePool` because we're appending to `filtered`, not `m.txpool.txns` directly.

This is probably the root cause of https://github.com/SiaFoundation/coreutils/issues/256. We had been blaming `parentMap` for that, so I ripped it out. I think it should stay ripped out, though, since benchmarks show that the operation it's optimizing is not particularly slow in the first place.

I also added an important validation step to `updateV2TransactionProofs`. This should prevent the panic seen in https://github.com/SiaFoundation/coreutils/pull/254.


lukechampine:
> [nitpick] Recomputing the entire parentMap on each call may degrade performance if this method is invoked frequently. Consider restoring a cache or documenting expected call frequency.

:omegalol:

lukechampine:
ok, slight snag here. The RHPv4 code does this in a few places:
```
- Fund transaction
- Call cm.V2TransactionSet to update the proofs
- Send transaction to host, receive host inputs
- Sign transaction
- Send signed to host
```
But this fails, because `V2TransactionSet` now fully validates the transaction, but the transaction is missing inputs and signatures.

I *think* we could solve this by not updating the proofs renter-side (since the host will update them anyway, after it has all the signatures). But that doesn't address the issue of `V2TransactionSet` on invalid inputs. So I think we should instead switch to validating only the Merkle proofs in `updateV2TransactionProofs`, e.g. by adding a `ValidateTransactionElements` method to `consensus.ElementAccumulator`.

n8maninger:
> ok, slight snag here. The RHPv4 code does this in a few places:
> 
> ```
> - Fund transaction
> - Call cm.V2TransactionSet to update the proofs
> - Send transaction to host, receive host inputs
> - Sign transaction
> - Send signed to host
> ```
> 
> But this fails, because `V2TransactionSet` now fully validates the transaction, but the transaction is missing inputs and signatures.
> 
> I _think_ we could solve this by not updating the proofs renter-side (since the host will update them anyway, after it has all the signatures). But that doesn't address the issue of `V2TransactionSet` on invalid inputs. So I think we should instead switch to validating only the Merkle proofs in `updateV2TransactionProofs`, e.g. by adding a `ValidateTransactionElements` method to `consensus.ElementAccumulator`.

We'd need to grab the txn parents some other way, but should be fine. I initially thought that might introduce a race with the contract elements, but it appears we rely on the host to fill it in already.

peterjan (approved):

PR #264, opened 2025-06-14
Timeout sync loop

n8maninger:
Adds a maximum amount of time we will sync with a single peer before trying another one. This is a hot fix for an issue seen on Mainnet where a single peer dominates the single threaded sync loop preventing peers from staying synced.


syncer/syncer.go
```diff
@@ -638,11 +642,7 @@ func (s *Syncer) peerLoop(ctx context.Context) error {
 			}
 
 			ctx, cancel := context.WithTimeout(ctx, s.config.ConnectTimeout)
-			if _, err := s.Connect(ctx, p); err != nil {
-				log.Debug("failed to connect to peer", zap.String("peer", p), zap.Error(err))
```
peterjan, line 642: def. not going to miss this one

peterjan (commented):
Looks good to me. Big change ... I remember Luke very purposefully restricting `peersForSync` to a handful of peers. I love it though, we should go for the robust option more often in these type of spots.

syncer/syncer.go
```diff
@@ -740,16 +734,29 @@ func (s *Syncer) syncLoop(ctx context.Context) error {
...
+				// would constantly accumulate defers, leaking memory until syncLoop
+				// returns.
+				resyncDelay := time.Duration(60+frand.Intn(60)) * time.Second
+				time.AfterFunc(resyncDelay, func() { p.setSynced(false) })
+			}
+
+			if newTip := s.cm.Tip(); newTip != oldTip {
 				s.log.Debug("finished syncing with peer", zap.Stringer("peer", p), zap.Stringer("newTip", newTip), zap.Uint64("blocks", totalBlocks))
 			} else {
 				s.log.Debug("finished syncing with peer, tip unchanged", zap.Stringer("peer", p), zap.Uint64("blocks", sentBlocks))
```
peterjan, line 756: we always log sent blocks but we should probably include the `SyncedBlocks` somewhere because that's some useful context we're missing at the moment

PR #265, opened 2025-06-16
chain: Don't ignore missing supplement in updateV2TransactionProofs

lukechampine:
The logic introduced in #193 caused a panic when a block supplement was missing: we would create an empty supplement, try to use it, then panic with out-of-range when trying to fetch stuff from the empty supplement.

In case you're wondering how we could end up in a situation where we have a block, but not a supplement: the supplement is only calculated when we fully validate and apply the block. If we receive a block that isn't on the best chain, we'll store it without also computing and storing the supplement. If someone then calls `V2TransactionSet` with a basis referencing that block, we'll attempt to use it.


chain/manager.go
```diff
@@ -1091,25 +1087,13 @@ func (m *Manager) updateV2TransactionProofs(txns []types.V2Transaction, from, to
 		return nil, fmt.Errorf("reorg path from %v to %v is too long (-%v +%v)", from, to, len(revert), len(apply))
 	}
 
-	var index types.ChainIndex
-	defer func() {
-		if panicErr := recover(); panicErr != nil {
```
peterjan, line 1096: I know we fixed a couple of spots now that avoid panicking, including this one, but should we get rid of the recover here? Feels like it doesn't hurt to keep it around for a bit.

PR #269, opened 2025-06-17
Increase max inflight RPC

n8maninger:
Increases the default number of RPCs that are allowed to be inflight concurrently to take into account the chattiness of Mainnet.


syncer/peer.go
```diff
@@ -484,7 +480,7 @@ func (s *Syncer) handleRPC(id types.Specifier, stream *gateway.Stream, origin *P
 			if err != nil {
 				s.log.Debug("received invalid transaction set", zap.Stringer("origin", origin), zap.Error(err))
 			} else {
-				s.relayV2TransactionSet(r.Index, r.Transactions, origin) // non-blocking
+				go s.relayV2TransactionSet(r.Index, r.Transactions, origin) // non-blocking
```
peterjan, line 483: This PR makes 3 relay actions non-blocking, but there's in total 4 of these (we made `relayV2BlockOutline` non-blocking in #264 ), all of these relay methods call `withV2Peers` outside of our threadgroup. I checked and I think it's probably fine, but we could consider adding them to the threadgroup.

lukechampine (dismissed):

PR #206, opened 2025-04-01
Satisfy elmo and fix ExpiringFileContracts ordering

lukechampine:

PR #271, opened 2025-06-18
chain: Preserve ID order in deleteFileContractExpiration

lukechampine:
We previously noticed that the order of expiring contract IDs was not stable under reorgs, since `putFileContractExpiration` always appended to the existing slice regardless of whether it was called during an apply or a revert. This was fixed in  https://github.com/SiaFoundation/coreutils/pull/206. However, a different form of the same bug remained: `deleteFileContractExpiration` was using the classic "swap-and-pop" trick to delete an element from within a slice; this is more efficient than shifting all the elements down, but it also changes their order. As a result, reverting and reapplying a block that deletes one ID from the middle of the slice is not idempotent. Consider the slice:
```
[A B C D]
```
If we delete A using swap-and-pop, we get:
```
[D B C]
```
If we then "undo" this by calling `putFileContractExpiration`, we'll just prepend, resulting in a different state:
```
[A D B C]
```
The correct approach is to delete via `s = append(s[:n], s[n+1:]...)`. Then, after deleting A, we have `[B C D]`, and prepending `A` results in the original slice.


peterjan (dismissed):
Nice!

ChrisSchinnerl (approved):

lukechampine:
Unfortunately this approach is untenable: `deleteFileContractExpiration` is called during application as well, so we can't modify its behavior without causing a huge fork.

So we're stuck with swap-and-pop, which is really annoying because swap-and-pop destroys information -- there's no way to undo it without tracking some extra state. I've added a new commit that does this: whenever we call `deleteFileContractExpiration`, we store the index of the deleted ID in a new bucket; then, during reverts, we call `restoreFileContractExpiration` (instead of `putFileContractExpiration`), and it looks up the index and inserts the ID in the right place.

You *could* run this without doing a rescan; as long as there isn't a reorg below the height when you started running it, you'll be ok. If you do encounter problems, you can resync from scratch and it should resolve. But we probably want to add a DB upgrade step that does a full rescan and computes the restoration index for all contracts. :/

n8maninger:
Instead of a full rescan, we can probably get away with populating it for the last 1008 or even all blocks since 526000.

lukechampine:
In theory there could be very old contracts that expire in the last month; better safe than sorry.

Added a migration that recomputes file contract elements + expirations from genesis.

lukechampine:
ok, this is ready to be tested on a synced consensus.db. I'm curious to see how long the migration takes.

PR #272, opened 2025-06-18
Prevent duplicate transactions when reverting mempool

n8maninger:


n8maninger:
Screenshot 2025-06-18 at 2 33 01 PM

Context

lukechampine:
yikes 👀

chain/manager.go
```diff
@@ -415,8 +415,21 @@ func (m *Manager) reorgTo(index types.ChainIndex) error {
 	m.txpool.medianFee = nil
 	if len(revert) > 0 {
 		b, _, _ := m.store.Block(revert[0].ID)
-		m.txpool.lastReverted = b.Transactions
-		m.txpool.lastRevertedV2 = b.V2Transactions()
+		m.txpool.lastReverted = m.txpool.lastReverted[:0]
+		// prevent coinbase transactions from spamming the pool
```
ChrisSchinnerl, line 419: (joke 😛) considering that we are already on the path to rustifying this codebase you might as well do 
```go
		m.txpool.lastReverted = slices.Collect(func(yield func(types.Transaction) bool) {
			for _, txn := range b.Transactions {
				if len(txn.MinerFees) == 0 {
					continue
				} else if !yield(txn) {
					return
				}
			}
		})
		m.txpool.lastRevertedV2 = slices.Collect(func(yield func(types.V2Transaction) bool) {
			for _, txn := range b.V2Transactions() {
				if txn.MinerFee.IsZero() {
					continue
				} else if !yield(txn) {
					return
				}
			}
		})
```

PR #275, opened 2025-06-23
Try file contract expirations

n8maninger:


peterjan:
Clever,  we could pretty easily parallelize it too. I'm not sure how valuable it is but we might want to write a small unit test around the permutations algorithm. E.g. not unit test our function per se but copy over the pattern in something that calculates string permutations or something.

peterjan:
This PR should probably also take along https://github.com/SiaFoundation/coreutils/pull/274/files#diff-f0a3575151db46e1e4391c0a1fdcb49402bcdb9e46b946491b2b6e5aa43814e1R645 ?

ChrisSchinnerl (commented):
Isn't this PR missing the fix in `deleteFileContractExpiration`? 

To make sure anyone who upgraded is on the correct state and can then reorg invalid blocks onto the correct state and to eventually have all nodes on the same state at the require height?

chain/manager.go
```diff
@@ -1321,6 +1355,49 @@ func (m *Manager) AddV2PoolTransactions(basis types.ChainIndex, txns []types.V2T
...
+func (m *Manager) ExpiringFileContractIDs(height uint64) []types.FileContractID {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.store.ExpiringFileContractIDs(height)
+}
+
+// OverwriteExpiringFileContractIDs overwrites the expiring file contract IDs at the given height.
+// This should not be called unless the IDs are known to be correct, as it will overwrite
+// any existing IDs at that height.
+func (m *Manager) OverwriteExpiringFileContractIDs(height uint64, ids []types.FileContractID) error {
```
peterjan, line 1368: This is currently not covered by the manager tests.

chain/manager.go
```diff
@@ -285,6 +289,30 @@ func (m *Manager) markBadBlock(bid types.BlockID, err error) {
...
+	}
+	bs.ExpiringFileContracts = bs.ExpiringFileContracts[:0]
+	for _, fcID := range order {
+		if fce, ok := elements[fcID]; ok {
+			bs.ExpiringFileContracts = append(bs.ExpiringFileContracts, fce)
+		} else {
+			return fmt.Errorf("missing expiring file contract %v in block %v", fcID, b.ID())
+		}
+	}
+	m.store.AddBlock(b, bs)
```
lukechampine, line 312: seems redundant?

chain/options.go
```diff
@@ -18,3 +21,13 @@ func WithPruneTarget(n uint64) ManagerOption {
...
 	}
 }
+
+// WithExpiringContractOrder sets the order of file contracts that are expiring
+// at a given height. This is used to work around a bug in the chain db
+// where the order of expiring file contracts is not preserved across
+// reorgs.
+func WithExpiringContractOrder(overwriteIDs map[types.BlockID][]types.FileContractID) ManagerOption {
+	return func(m *Manager) {
+		m.expiringFileContractOrder = overwriteIDs
```
lukechampine, line 31: do we need this to be configurable?

chain/manager.go
```diff
@@ -309,7 +337,9 @@ func (m *Manager) applyTip(index types.ChainIndex) error {
 	} else if bs == nil {
 		bs = new(consensus.V1BlockSupplement)
 		*bs = m.store.SupplementTipBlock(b)
-		if err := consensus.ValidateBlock(m.tipState, b, *bs); err != nil {
+		if err := m.overwriteExpirations(b, bs); err != nil {
```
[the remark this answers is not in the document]
n8maninger: @lukechampine can it? I was assuming if one of these blocks already has a supplement on disk and needed to be reverted it must use the existing block supplement, not the overwritten one. If not, then yeah that's much cleaner.

lukechampine:
Also need to call `overwriteProofs` in `updateV2TransactionProofs`

chain/manager.go
```diff
@@ -1131,6 +1163,8 @@ func (m *Manager) updateV2TransactionProofs(txns []types.V2Transaction, from, to
 			return nil, fmt.Errorf("missing applied block at index %v", index)
 		} else if bs == nil {
 			return nil, fmt.Errorf("missing applied block supplement at index %v", index)
+		} else if err := m.overwriteExpirations(b, bs); err != nil {
```
ChrisSchinnerl, line 1166: I wonder if we could move this into `blockAndChild` and `blockAndParent`? That seems to catch all spots where we need to call `overwriteExpirations`?

peterjan (approved):

lukechampine (approved):

PR #290, opened 2025-07-09
Run fuzzer after PRs are merged

chris124567:
The same workflow has been tested at https://github.com/SiaFoundation/fuzz/actions

The default mines 500 blocks with allow height at 250 and require height at 400.  For each random block we generate, we apply it, revert it, then apply again.  We check that the expiring contracts after the applications and before application/after revert are the same (we sort ExpiringFileContracts by ID before comparing).  The generated blocks are uploaded as an artifact for reproducibility (you can reproduce with `./fuzzer repro repro.json`).


.github/workflows/fuzzer.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  fuzzer:
+    name: Fuzzer
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Configure git
+        run: git config --global core.autocrlf false # required on Windows
```
n8maninger, line 16: This step is only necessary for golangci-lint on Windows

.github/workflows/fuzzer.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/checkout@v4
+        with:
+          repository: SiaFoundation/fuzz
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Build
+        run: go get go.sia.tech/core@none && go get go.sia.tech/coreutils@master && go mod tidy && go build -buildvcs=false .
```
n8maninger, line 27: `@master` will probably always work, but we've noticed annoying caching in other workflows. Probably better to explicitly grab the commit hash you are wanting to test. Should be able to use `@$GITHUB_SHA`.

n8maninger (changes requested):

PR #297, opened 2025-07-17
Fix RPCError code and add `DeleteSector`

peterjan:
Testing data integrity in `indexd` I noticed the `RPCError` returned by RPCVerifySector has a wrong code. I also extended the `EphemeralSectorStore` with a `DeleteSector` to enable mocking a host that lost it.

### DIFF
--- a/rhp/v4/server.go
+++ b/rhp/v4/server.go
@@ -1112,13 +1112,19 @@ func (s *Server) handleRPCVerifySector(stream net.Conn) error {
 	}
 	prices, token := req.Prices, req.Token
 
+	if exists, err := s.sectors.HasSector(req.Root); err != nil {
+		return fmt.Errorf("failed to check sector: %w", err)
+	} else if !exists {
+		return rhp4.ErrSectorNotFound
+	}
+
 	if err := s.contractor.DebitAccount(token.Account, prices.RPCVerifySectorCost()); err != nil {
 		return fmt.Errorf("failed to debit account: %w", err)
 	}
 
 	sector, err := s.sectors.ReadSector(req.Root)
 	if err != nil {
-		return rhp4.NewRPCError(rhp4.ErrorCodeHostError, err.Error())
+		return fmt.Errorf("failed to read sector: %w", err)
 	}
 
 	proof := rhp4.BuildSectorProof(sector, req.LeafIndex, req.LeafIndex+1)
--- a/rhp/v4/server.go
+++ b/rhp/v4/server.go
@@ -1118,7 +1118,7 @@ func (s *Server) handleRPCVerifySector(stream net.Conn) error {
 
 	sector, err := s.sectors.ReadSector(req.Root)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


testutil/host.go
```diff
@@ -56,6 +56,17 @@ type EphemeralSectorStore struct {
 
 var _ rhp4.Sectors = (*EphemeralSectorStore)(nil)
 
+// DeleteSector deletes a sector from the store.
+func (es *EphemeralSectorStore) DeleteSector(root types.Hash256) error {
+	es.mu.Lock()
+	defer es.mu.Unlock()
+	if _, ok := es.sectors[root]; !ok {
+		return errors.New("sector not found")
```
n8maninger, line 64: Might as well return `proto4.ErrSectorNotFound`

rhp/v4/server.go
```diff
@@ -1107,7 +1107,7 @@ func (s *Server) handleRPCVerifySector(stream net.Conn) error {
 
 	sector, err := s.sectors.ReadSector(req.Root)
 	if err != nil {
-		return rhp4.NewRPCError(rhp4.ErrorCodeBadRequest, err.Error())
+		return rhp4.NewRPCError(rhp4.ErrorCodeHostError, err.Error())
 	}
```
ChrisSchinnerl, line 1111: Couldn't this be `return err`? Either `ReadSector` already returns `ErrSectorNotFound` in which case the status code is `ErrorCodeHostError` or it returns some arbitrary error in which case the caller will return a `ErrHostInternalError`. 

Regardless I got one thing that might be good to double check. The integration test shouldn't really care about the error returned here. Both errors should result in the integrity check failing since we don't want to rely on specific errors returned by the host so that they can't swindle themselves out of integration checks.

peterjan: Hm ok so in `handleRPCReadSector` we check if it exists first. I think the most correct and consistent impl. would be to perform this check as well, return ` rhp4.ErrSectorNotFound` and indeed return the plain `err` in case the read fails.

I'll make sure to double check the integration tests in indexd. We do currently check for the error that includes the error code. Having the error code be part of the string is quite error prone if you check for `strings.Contains`...


```
if exists, err := s.sectors.HasSector(req.Root); err != nil {
    return fmt.Errorf("failed to check sector: %w", err)
} else if !exists {
    return rhp4.ErrSectorNotFound
}
```

rhp/v4/server.go
```diff
@@ -1101,13 +1101,19 @@ func (s *Server) handleRPCVerifySector(stream net.Conn) error {
...
 	prices, token := req.Prices, req.Token
 
+	if exists, err := s.sectors.HasSector(req.Root); err != nil {
```
ChrisSchinnerl, line 1104: @n8maninger is it safe to debit the account after checking if the sector exists? Seems like it opens up the host to renters repeatedly forcing it to perform disk I/O without payment. Same goes for `RPCReadSector`.

ChrisSchinnerl (approved):
Apart from the open question to Nate it looks good to me.

PR #311, opened 2025-08-11
Increase RHP4 RPC timeout

ChrisSchinnerl:
Some people are reporting uploading issues after RHP4 likely related to timeouts being too low now.
This increases the timeout to 5 minutes, to match the previous RHP3 timeout in `handleRPCExecute`.

Related to https://github.com/SiaFoundation/renterd/issues/1948#issuecomment-3172784475

### DIFF
--- a/.changeset/increase_rhp4_rpc_timeout_to_5_minutes.md
+++ b/.changeset/increase_rhp4_rpc_timeout_to_5_minutes.md
@@ -2,4 +2,4 @@
 default: patch
 ---
 
-# Increase default RHP4 RPC timeout to 5 minutes.
+# Increase default RHP4 RPC timeout to 3 minutes.
--- a/.changeset/increase_rhp4_rpc_timeout_to_5_minutes.md
+++ b/.changeset/increase_rhp4_rpc_timeout_to_5_minutes.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Increase default RHP4 RPC timeout to 5 minutes.
--- a/rhp/v4/server.go
+++ b/rhp/v4/server.go
@@ -1150,7 +1150,7 @@ func (s *Server) handleHostStream(stream net.Conn, log *zap.Logger) {
 		}
 	}()
 
-	stream.SetDeadline(time.Now().Add(5 * time.Minute)) // RPC timeout
+	stream.SetDeadline(time.Now().Add(3 * time.Minute)) // RPC timeout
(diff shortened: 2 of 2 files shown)


n8maninger:
2-3m would be more reasonable 5 is a bit high imo. That’s 0.1Mbps transfer speed. Timing out is probably the better choice.

jdannenberg:
> 2-3m would be more reasonable. 5 is a bit high imo. That’s 0.1Mbps transfer speed. Timing out is probably the better choice.

Is that the case, really? Assuming only one 120MiB slab at a time (no overdrive,etc) to take 5 Minutes to upload, upload speed is 120MiB / 300seconds ~ 0,4MiB/s ..... so 3,2 Mbps, not 0,1Mbps? We cannot really discuss this on a per sector timeout basis here since usually all sectors in a slab share the same bandwidth, can we?

Am I missing something? 3,2Mbps is not that slow especially considering that there might be parallel slab uploads.

n8maninger:
> Is that the case, really? Assuming only one 120MiB slab at a time (no overdrive,etc) to take 5 Minutes to upload, upload speed is 120MiB / 300seconds ~ 0,4MiB/s ..... so 3,2 Mbps, not 0,1Mbps? We cannot really discuss this on a per sector timeout basis here since usually all sectors in a slab share the same bandwidth, can we?

This is on the host side, so timeouts need to be considered per sector. Each sector upload ties up host resources independently, regardless of slab bandwidth sharing on the renter. If a renter is bandwidth-limited, they have other options, like reducing parallelism. Locking up host resources for longer because a renter may not have the bandwidth isn't ideal.

jdannenberg:
> This is on the host side, so timeouts need to be considered per sector. Each sector upload ties up host resources independently, regardless of slab bandwidth sharing on the renter.

I find it difficult to agree to that one because ultimately every sector upload of a slab is started at the very same time and thus they all do share bandwidth.

> There are better options for tuning for low bandwidth on the renter side that don't require locking up host resources (i.e. reducing parallelism).

And this one is easy to agree with, I see why we are now at 3 Minutes timeout. However, renter is lacking the functionality of reducing sector upload parallelism. Only slab parallelism is configurable, but for a single slack you already need at least 30-40mbit free upload bandwidth.

Maybe sector parallelism should be configurable in renterd indeed. I implemented a somewhat uglier solution here: https://github.com/SiaFoundation/renterd/pull/1976

PR #316, opened 2025-08-15
Reset Zen testnet

ChrisSchinnerl:
Due to a third party mining a lot of blocks in a short time on the usually CPU-mined Zen testnet, it is stuck. This PR resets it.

### DIFF
--- a/.changeset/reset_zen_testnet.md
+++ b/.changeset/reset_zen_testnet.md
@@ -1,0 +1,5 @@
+---
+default: major
+---
+
+# Reset Zen testnet.


chain/network.go
```diff
@@ -86,7 +86,7 @@ func Mainnet() (*consensus.Network, types.Block) {
 		}{
 			AllowHeight:    526000, // June 6th, 2025 @ 6:00am UTC
 			RequireHeight:  530000, // July 4th, 2025 @ 2:00am UTC
-			FinalCutHeight: 600000,
+			FinalCutHeight: 552100, // December 1st, 2025 @ 11:00am UTC
```
ChrisSchinnerl, line 89: Please double check the math. I did round it down since it won't be accurate anyway but it should be in the ballpark.

lukechampine:
```
dec_1 - aug_15 = 108 days
current_height = 536570
(108 * 144) + 536570 = 552122 
```
alternative check:
```
552100 - 530000 = 22100
22100 * 10 minutes = 153 days
jul_4 + 153 days = dec_4
```
lgtm. A 3 day spread is obviously not ideal, but it's tolerable.

mike76-dev:
Would it be feasible to implement a nonce denominator for Zen, similar to the one for the ASIC hardfork, to prevent Zen from being ASIC-mineable?

lukechampine:
clever! Yes, we could set `NonceFactor` to some large multiple of `1009`. Somewhere around 100,000 * 1009 ought to do it.

lukechampine:
(I support this mainly because it retroactively justifies my decision to make `NonceFactor` a method of `State` :P)

mike76-dev:
Zen hasn't been operational for more than a week. People can't use it now. I'm waiting myself to get it back up to test my code. It's going to be a hassle again and again if someone decides to point their ASIC at Zen. If it's easy enough to protect against that (and I understand it is), let's do it.

n8maninger:
Feel free to make a PR to `core` adding `NonceFactor` to the `HardforkASIC` struct and using that as the return value for `NonceFactor` in the ConsensusState struct.

PR #320, opened 2025-08-25
Set keepalive period for siamux TCP connections

ChrisSchinnerl:


rhp/v4/siamux/siamux.go
```diff
@@ -134,6 +155,11 @@ func Serve(l net.Listener, s *rhp4.Server, log *zap.Logger) {
 		}
 		log := log.With(zap.Stringer("peerAddress", conn.RemoteAddr()))
 
+		if tcpConn, ok := conn.(*net.TCPConn); ok && so.TCPKeepalivePeriod > 0 {
+			tcpConn.SetKeepAlive(true)
+			tcpConn.SetKeepAlivePeriod(30 * time.Second)
+		}
```
ChrisSchinnerl, line 161: I have not been able to test this yet. Seems like the only way to do that (since it happens on the OS level) would be a manual test where we kill the process.

PR #323, opened 2025-08-25
Configurable RPC timeout

n8mgr:


rhp/v4/siamux/siamux.go
```diff
@@ -158,7 +158,7 @@ func Serve(l net.Listener, s *rhp4.Server, log *zap.Logger, opts ...ServeOption)
 
 		if tcpConn, ok := conn.(*net.TCPConn); ok && so.TCPKeepalivePeriod > 0 {
 			tcpConn.SetKeepAlive(true)
-			tcpConn.SetKeepAlivePeriod(30 * time.Second)
+			tcpConn.SetKeepAlivePeriod(so.TCPKeepalivePeriod)
```
ChrisSchinnerl, line 161: Noticed this during my testing. Good news is that the server gets unblocked with a `peer closed underlying connection`.

PR #331, opened 2025-09-12
Implement parallel syncing

lukechampine:
Replaces the old serial syncing algorithm a fancy parallel one. In brief:

```
Old algorithm:
1. For each unsynced peer, request blocks until the peer reports no further blocks

New algorithm:
1. In parallel, for each unsynced peer, request up to 10k headers
2. In parallel, request sections of the header chain from peers in batches of 100 blocks
3. Poll for new peers in the background and immediately assign them work
4. After assigning the final batch, redundantly assign any remaining in-progress batches to idle peers
5. Add batches of blocks to the Manager in a separate goroutine to avoid blocking peer I/O
6. When requesting v2 blocks, perform validation within each worker goroutine, rather than in the Manager goroutine
```

Empirically, the new algorithm is anywhere from 2x-10x faster, particularly once it reaches the v2 require height and can start validating in parallel as well.

### DIFF
--- a/wallet/wallet_test.go
+++ b/wallet/wallet_test.go
@@ -1925,7 +1925,7 @@ func TestRebroadcastTransaction(t *testing.T) {
 	// create wallet
 	l := zaptest.NewLogger(t)
 	s := &testutil.MockSyncer{}
-	w, err := wallet.NewSingleAddressWallet(pk, cm, ws, s, wallet.WithLogger(l.Named("wallet")), wallet.WithDebounceInterval(time.Millisecond))
+	w, err := wallet.NewSingleAddressWallet(pk, cm, ws, s, wallet.WithLogger(l.Named("wallet")), wallet.WithDebounceInterval(50*time.Millisecond))
 	if err != nil {
 		t.Fatal(err)
 	}


n8mgr:
🧪

chain/manager.go
```diff
@@ -298,6 +301,55 @@ func (m *Manager) AddBlocks(blocks []types.Block) error {
...
+		if blocks[i].V2 == nil {
+			return fmt.Errorf("only v2 blocks can be pre-validated")
+		}
+		m.store.AddBlock(blocks[i], &consensus.V1BlockSupplement{})
+		m.store.AddState(states[i])
+	}
+
+	// if this chain is now the best chain, trigger a reorg
+	cs := states[len(states)-1]
+	if cs.SufficientlyHeavierThan(m.tipState) {
```
ChrisSchinnerl, line 326: Everything below this line can probably be extracted into a `triggerReorg` helper.

chain/manager.go
```diff
@@ -298,6 +301,55 @@ func (m *Manager) AddBlocks(blocks []types.Block) error {
...
 
+// AddValidatedV2Blocks ingests a chain of v2 blocks. The blocks must already be
+// validated, and the first block's parent must be known. If the chain has
+// sufficient work, it may become the new best chain, triggering a reorg.
+func (m *Manager) AddValidatedV2Blocks(blocks []types.Block, states []consensus.State) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if len(blocks) == 0 {
+		return nil
+	}
```
ChrisSchinnerl, line 312: should probably compare the length of `blocks` and `states` here to avoid panics further down.

syncer/parallel_sync.go
```diff
@@ -0,0 +1,201 @@
...
+			resp.blocks = blocks
+			for _, b := range blocks {
+				if err := consensus.ValidateBlock(cs, b, consensus.V1BlockSupplement{}); err != nil {
+					return Resp{req: req, err: err}
+				}
+				cs, _ = consensus.ApplyBlock(cs, b, consensus.V1BlockSupplement{}, time.Time{})
+				resp.states = append(resp.states, cs)
+			}
+		} else {
+			blocks, _, err := p.SendV2Blocks(ctx, []types.BlockID{req.base.ID}, req.numBlocks, s.config.SendBlocksTimeout)
```
ChrisSchinnerl, line 75: Not sure I follow the logic here. What about v1 blocks? At this point it seems like the height could be before the allow height?

syncer/parallel_sync.go
```diff
@@ -0,0 +1,201 @@
...
+			for i < len(resps) && resps[i] != nil {
+				r := resps[i]
+				var err error
+				if r.req.base.Height >= cs.Network.HardforkV2.RequireHeight {
+					err = s.cm.AddValidatedV2Blocks(r.blocks, r.states)
+				} else {
+					err = s.cm.AddBlocks(r.blocks)
+				}
+				if err != nil {
+					errCh <- err
```
ChrisSchinnerl, line 121: Can this ever happen outside of a persistence error? My concern here is that an invalid block breaks syncing.

chain/manager.go
lukechampine: eh, it gets kinda ugly due to the onReorg functions needed to be called without a lock; `triggerReorg` would be a weird function that assumes it is called under lock, but then unlocks in the middle? Plus the name ambiguity of `m.reorgTo` vs `m.triggerReorg`. Not worth it imo

syncer/parallel_sync.go
lukechampine: yeah, not the best name. In this case `V2` refers to the v2 *encoding* of blocks. The blocks themselves can be v1 blocks.

Now that the old `SendBlocks` is deprecated, we could rename the method to just `SendBlocks`.

syncer/syncer_test.go
```diff
@@ -87,6 +87,38 @@ func TestSyncer(t *testing.T) {
 	}
 }
 
+func TestParallelSync(t *testing.T) {
```
ChrisSchinnerl, line 90: Isn't this just a portion of copy pasta from `TestSyncer` with more blocks being mined to get past the require height? Might as well just delete `TestSyncer` and name this `TestSyncer` instead?

The kind of testing I miss is the not so happy path kind. Where we sync the most recent blocks successfully for example and then the earlier blocks fail and then check that our node isn't completely screwed. 
Maybe then connect to another syncer that has the earlier blocks and see if that finishes the sync successfully.

lukechampine:
Ok, added a real test. This indeed caught a few issues.

chain/manager.go
```diff
@@ -298,9 +303,63 @@ func (m *Manager) AddBlocks(blocks []types.Block) error {
...
+// AddValidatedV2Blocks ingests a chain of v2 blocks. The blocks must already be
+// validated, and the first block's parent must be known. If the chain has
+// sufficient work, it may become the new best chain, triggering a reorg.
+func (m *Manager) AddValidatedV2Blocks(blocks []types.Block, states []consensus.State) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if len(blocks) == 0 {
+		return nil
+	} else if len(states) != len(blocks) {
+		panic("chain: expected same number of blocks and states") // developer error
```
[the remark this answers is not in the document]
lukechampine: I would happily call your bluff and turn `only v2 blocks can be pre-validated` into a panic, but I will play nice :P

chain/manager.go
```diff
@@ -298,9 +303,63 @@ func (m *Manager) AddBlocks(blocks []types.Block) error {
...
+		}
+		m.mu.Lock()
+	}
+	return nil
+}
+
 // markBadBlock marks a block as bad, so that we don't waste resources
 // re-validating it if we see it again.
 func (m *Manager) markBadBlock(bid types.BlockID, err error) {
+	if strings.Contains(err.Error(), "supplement") || strings.Contains(err.Error(), "commitment") {
```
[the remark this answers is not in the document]
lukechampine: We should probably just strip out the `badBlock` system entirely

syncer/parallel_sync.go
lukechampine: immediately after the `SendV2Blocks` calls

syncer/peer.go
```diff
@@ -168,9 +199,9 @@ func (p *Peer) RelayV2TransactionSet(index types.ChainIndex, txns []types.V2Tran
 // SendV2Blocks requests up to n blocks from p, starting from the most recent
 // element of history known to p. The peer also returns the number of remaining
 // blocks left to sync.
-func (p *Peer) SendV2Blocks(history []types.BlockID, maxBlocks uint64, timeout time.Duration) ([]types.Block, uint64, error) {
+func (p *Peer) SendV2Blocks(ctx context.Context, history []types.BlockID, maxBlocks uint64, timeout time.Duration) ([]types.Block, uint64, error) {
```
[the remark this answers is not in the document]
lukechampine: maybe later

peterjan (dismissed):

n8mgr (approved):

PR #365, opened 2025-11-26
Proposal: Use OpenStreamSync instead of OpenStream for QUIC

ChrisSchinnerl:
I saw a lot of "too many open streams" errors in `indexd` after merging the new client. Turns out quic defaults to a limit of 100 incoming streams which we set to 1000 but we still run into it now that migrations and sector verifications use the new client.  

Funny enough sector verifications started to print logs like 999 successful verifications and 1 failed ones. Almost like streams wouldn't get closed fast enough.

⚠️ unfortunately this isn't fixing the actual speed issue but will just highlight them in a different way. Instead of the "too many open streams" error we'll get uploads that time out after 2 minutes.


n8mgr:
Well that’s nice. I noticed when I was adding the QUIC proto that stream closing is slow even if you’re closing it on your side. I figured it would be one ack packet and be done, but doesn’t seem to be the case.

PR #366, opened 2025-11-26
Add final Read when closing QUIC stream + reproduction test

ChrisSchinnerl:
This PR adds `CancelRead` to `stream.Close`. You might wonder why that is necessary and I'm glad you ask.

Here is some documentation:  https://quic-go.net/docs/quic/streams/#bidirectional-stream

I can already tell you are still not convinced so I included a simple test in this PR. `TestStreamLimit`.
With `s.CancelRead` the test passes. Without it, the test blocks forever.

However, the part that makes this a bit more annoying is that both parties need to call `s.CancelRead` for the test to pass.
It is not enough for the client to do so. You can verify that by deleting the line `transport := &transport{conn}` in the test and instead use the raw quic connection to avoid the wrapped `Close`.

TLDR; This fixes the issue but hosts need to upgrade too

### DIFF
--- a/rhp/v4/quic/quic.go
+++ b/rhp/v4/quic/quic.go
@@ -29,7 +29,7 @@ const (
 
 	// maxIncomingStreams is the maximum number of incoming streams allowed per
 	// QUIC connection by the server.
-	maxIncomingStreams = 1000
+	maxIncomingStreams = 100000
 )
 
 type (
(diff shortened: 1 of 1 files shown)


rhp/v4/rpc_test.go
```diff
@@ -1276,7 +1276,7 @@ func TestRPCTimeout(t *testing.T) {
 
 	isTimeoutErr := func(err error) bool {
 		t.Helper()
-		return err != nil && (strings.Contains(err.Error(), "deadline exceeded") || strings.Contains(err.Error(), "i/o timeout"))
+		return err != nil && (strings.Contains(err.Error(), "deadline exceeded") || strings.Contains(err.Error(), "i/o timeout") || strings.Contains(err.Error(), "canceled by local with error code 1"))
```
ChrisSchinnerl: This is annoying. We keep accumulating error that might happen when we reach a timeout. So we must not forget to update `shouldResetTransport`. This error is effectively the "stream was gracefully closed" error of QUIC.

peterjan (dismissed):

Issue #391, opened 2026-02-03
WebTransport listener is misconfigured


Alrighttt:
While experimenting with Rust WASM tooling, I've discovered that the WebTranport listening is seemingly broken. 

CONNECT was failing, so I had to use the `webtransport.ConfigureHTTP3Server()` method:
```
	h3Server := &http3.Server{
		Handler: mux,
	}

	// Configure the HTTP/3 server for WebTransport (enables extended CONNECT, etc.)
	webtransport.ConfigureHTTP3Server(h3Server)

	wts := &webtransport.Server{
		H3: h3Server,
		CheckOrigin: func(*http.Request) bool {
			return true // allow all origins
		},
		// no need for TLS config since it will already be negotiated
	}
```
https://github.com/SiaFoundation/coreutils/blob/ff3cac4a06bcb226d041e46c88400c687407aaec/rhp/v4/quic/quic.go#L271

This was still failing the handshake, so I debugged further. We weren't logging the error when calling `wts.ServeQUICConn()`, so I added:
```
				err := wts.ServeQUICConn(conn)
				if err != nil {
					log.Debug("ServeQUICConn returned with error", zap.Error(err))
				}

```
https://github.com/SiaFoundation/coreutils/blob/ff3cac4a06bcb226d041e46c88400c687407aaec/rhp/v4/quic/quic.go#L314

Which prints:
`2026-02-03T11:37:15-05:00       DEBUG   rhp4.quic       ServeQUICConn returned with error{"peerAddress": "127.0.0.1:50221", "protocol": "h3", "error": "webtransport: QUIC Stream Resets with Partial Delivery required, enable it via QUICConfig.EnableStreamResetPartialDelivery"}`

Finally, I enabled this:
```
&quic.Config{
		EnableDatagrams:                  true,
		EnableStreamResetPartialDelivery: true,
		KeepAlivePeriod:                  30 * time.Second,
		MaxIdleTimeout:                   30 * time.Minute,
		MaxIncomingStreams:               maxIncomingStreams,
	}
```
https://github.com/SiaFoundation/coreutils/blob/ff3cac4a06bcb226d041e46c88400c687407aaec/rhp/v4/quic/quic.go#L226

I can now connect via a javascript console:
```
const url = 'https://<snip>.sia.host:9984/sia/rhp/v4';
const wt = new WebTransport(url);
wt.ready.then(() => console.log('Connected!')).catch(e => console.error('Error:', e.message));
[...]

PR #392, opened 2026-02-03
Fix webtransport conn not being served due to missing `EnableStreamResetPartialDelivery`

ChrisSchinnerl:
Seems like a new requirement for `webtransport-go`. 

Relevant code can be found  [here](https://github.com/quic-go/webtransport-go/commit/9d448b125754f4c83064afb2c586221214e55eec).

Closes https://github.com/SiaFoundation/coreutils/issues/391


n8mgr (commented):
Do we need to call `webtransport.ConfigureHTTP3Server` as described in the issue?

chris124567 (approved):

PR #401, opened 2026-02-23
validate RPCFundAccountsRequest

Alrighttt:
This method was implemented but never used.


ChrisSchinnerl (approved):
Pushed an additional call to .Validate on the server side

PR #421, opened 2026-04-28
Add RHP4 pool support for shared balance pools

n8mgr:
Hosts can now back accounts with shared balance pools so renters don't need to pre-fund and continually rebalance every account. A renter funds a pool once and attaches as many accounts as they want; debits drain the account's own balance first and fall through to attached pools, which keeps per-account allowances small and reduces the total capital sitting idle in account balances. This is the host-side companion to the new RHP4 pool RPCs.


testutil/host.go
```diff
@@ -448,6 +576,8 @@ func NewEphemeralContractor(cm *chain.Manager) *EphemeralContractor {
 		roots:            make(map[types.FileContractID][]types.Hash256),
 		locks:            make(map[types.FileContractID]bool),
 		accounts:         make(map[proto4.Account]types.Currency),
+		pools:            make(map[proto4.Account]types.Currency),
+		attached:         make(map[proto4.Account][]proto4.Account),
```
chris124567, line 580: I know this is just the testutil host and the real host would store it on disk but this made me wonder if there should be any limit on the number of accounts per pool or if we should just enforce that in hostd.  We could also make RPCAttachPools cost something

rhp/v4/rpc.go
```diff
@@ -815,6 +842,199 @@ func RPCReplenishAccounts(ctx context.Context, t TransportClient, p RPCReplenish
...
+	if err := req.Validate(); err != nil {
+		return RPCFundPoolsResult{}, fmt.Errorf("invalid request: %w", err)
+	}
+
+	var resp rhp4.RPCFundAccountsResponse
+	if err := callSingleRoundtripRPC(ctx, t, rhp4.RPCFundPoolsID, &req, &resp); err != nil {
+		return RPCFundPoolsResult{}, err
+	}
+
+	if len(resp.Balances) != len(deposits) {
```
peterjan: `nit` should we add a similar length check in `RPCReplenishPools`?

rhp/v4/rpc.go
```diff
@@ -815,6 +842,199 @@ func RPCReplenishAccounts(ctx context.Context, t TransportClient, p RPCReplenish
...
+	}
+
+	for _, deposit := range resp.Deposits {
+		if deposit.Amount.Cmp(p.Target) > 0 {
+			return RPCReplenishPoolsResult{}, fmt.Errorf("expected deposit <= %v, got %v", p.Target, deposit.Amount)
+		}
+	}
+
+	totalCost := resp.TotalCost()
+	if totalCost.IsZero() {
```
peterjan, line 883: `nit` we could add a test case that asserts the contract is not revised when this is the case, effectively testing this early return

peterjan (commented):
Left two small nits but otherwise LGTM

testutil/host.go
```diff
@@ -298,16 +302,140 @@ func (ec *EphemeralContractor) CreditAccountsWithContract(deposits []proto4.Acco
...
+
+// AttachPools applies a batch of attachments atomically. All referenced
+// pools must exist before any mutation occurs.
+func (ec *EphemeralContractor) AttachPools(attachments []proto4.PoolAttachment) error {
+	ec.mu.Lock()
+	defer ec.mu.Unlock()
+
+	for _, a := range attachments {
+		if _, ok := ec.pools[a.Pool]; !ok {
+			return fmt.Errorf("pool %v does not exist", a.Pool)
```
ChrisSchinnerl: Would be nice to provide a proper RPC error for this and the hostd implementation to get something better than an internal error in case you try to attach before funding a pool.

testutil/host.go
n8mgr: We don't have any limits on accounts *yet*, but I agree it's not ideal and an easy path to abuse. Feel free to tackle it in `hostd`.

rhp/v4/rpc.go
ChrisSchinnerl, line 38: We need a new version here so that indexers know whether it's ok to use the new RPCs.

rhp/v4/server.go
ChrisSchinnerl, line 1: The `protocolVersion` constant in this file needs to be updated to whatever the new current one is.

rhp/v4/server.go
```diff
@@ -118,8 +118,24 @@ type (
...
+
+		// PoolBalances returns balances for the given pool keys in input
+		// order. Unknown pools return types.ZeroCurrency.
+		PoolBalances([]rhp4.Account) ([]types.Currency, error)
+		// CreditPoolsWithContract atomically revises a contract and credits
+		// the pools. Pools auto-create on first credit.
+		CreditPoolsWithContract([]rhp4.AccountDeposit, types.FileContractID, types.V2FileContract, rhp4.Usage) ([]types.Currency, error)
+		// AttachPools applies a batch of attachments atomically. Each pool
+		// must already exist; idempotent on existing attachments.
+		AttachPools([]rhp4.PoolAttachment) error
```
ChrisSchinnerl, line 135: (nit) In the ephemeral contractor implementation the attach order impacts the spend order. Which is a very nice property to have if we ever wanted to do something like a shared kickstarting pool for all users to reduce the time it takes to fund their accounts. That said there is currently no way to change the pool order without detaching a pool first.

So it would be nice if attaching an already attached pool again would place it to the end of the list. To allow for reordering pools without unlinking them first.

e.g. I have pools A, B and C attached and call attach again with C and A, I'd end up with the order B, C and A. So the relative order matches the RPC call and any pool that hasn't been part of the RPC will end up with a higher priority.


I'd also mention in the comment that the attachment order matters

ChrisSchinnerl (changes requested):

rhp/v4/rpc.go
```diff
@@ -815,6 +835,149 @@ func RPCReplenishAccounts(ctx context.Context, t TransportClient, p RPCReplenish
...
+			Account:    in.Account,
+			Pool:       rhp4.Account(in.PoolKey.PublicKey()),
+			ValidUntil: deadline,
+		}
+		a.Signature = in.PoolKey.SignHash(a.SigHash(hostKey))
+		attachments = append(attachments, a)
+	}
+	req := rhp4.RPCAttachPoolsRequest{Attachments: attachments}
+	if err := req.Validate(); err != nil {
+		return fmt.Errorf("invalid request: %w", err)
```
peterjan, line 955: shouldn't we return `rhp4.NewRPCError(rhp4.ErrorCodeBadRequest, err.Error())`

rhp/v4/rpc.go
```diff
@@ -815,6 +835,149 @@ func RPCReplenishAccounts(ctx context.Context, t TransportClient, p RPCReplenish
...
+			Account:    in.Account,
+			Pool:       in.Pool,
+			ValidUntil: deadline,
+		}
+		d.Signature = in.Signer.SignHash(d.SigHash(hostKey))
+		detachments = append(detachments, d)
+	}
+	req := rhp4.RPCDetachPoolsRequest{Detachments: detachments}
+	if err := req.Validate(); err != nil {
+		return fmt.Errorf("invalid request: %w", err)
```
peterjan, line 978: shouldn't we return `rhp4.NewRPCError(rhp4.ErrorCodeBadRequest, err.Error())`

.changeset/add_rhp4_pool_rpc_handlers_and_renter_helpers.md
```diff
@@ -0,0 +1,9 @@
+---
+default: major
+---
+
+# Add RHP4 pool RPC handlers, renter helpers, and Contractor interface methods for shared balance pools.
+
+Hosts can now back accounts with shared balance pools so renters don't need to pre-fund and continually rebalance every account. A renter funds a pool once and attaches as many accounts as they want; debits drain the account's own balance first and fall through to attached pools, which keeps per-account allowances small and reduces the total capital sitting idle in account balances. This is the host-side companion to the new RHP4 pool RPCs.
+
+Extends the RHP4 `Contractor` interface with `PoolBalances`, `CreditPoolsWithContract`, `AttachPools`, and `DetachPools` (existing implementations must be updated). Adds host-side handlers for the new pool RPCs and renter-side helpers `RPCFundPools`, `RPCReplenishPools`, `RPCAttachPools`, and `RPCDetachPools`. `DebitAccount` semantics are extended to drain the account's own balance first, then attached pools in attachment order.
```
peterjan: There's no `RPCFundPools`

peterjan (changes requested):

rhp/v4/rpc.go
n8mgr: Validate returns a request error

rhp/v4/rpc.go
n8mgr: Validate returns an RPC error already

peterjan (approved):

ChrisSchinnerl (approved):

PR #424, opened 2026-05-05
Check resp.DataLength before verifying proof

ChrisSchinnerl:
The proof verification is going to fail if we didn't download the full data so we should return a length mismatch first since that's going to be more descriptive than "invalid proof".


chris124567 (approved):

peterjan (approved):
Was going to suggest adding a 🧪 but it's a little cumbersome indeed. You'd have to mock a transport and have a server advertise a certain number of bytes but only send half, doesn't hurt to add it though.

PR #437, opened 2026-06-05
Read concurrency

Alrighttt:
Changes `Manager.mu` to `sync.RWMutex` to enable reads to run in parallel.


Alrighttt:
As discussed with Nate and Luke, I will be redesigning or removing the peer cancellation logic. Don't bother with review for now.

syncer/peer.go
```diff
@@ -24,8 +24,16 @@ type Peer struct {
 	mu       sync.Mutex
 	synced   bool
 	err      error
+	// goroutines spawned for this peer can observe ctx.Done() to short-circuit when the peer is
+	// gone.
+	ctx context.Context
```
n8mgr: Go advises against keeping contexts on the struct because it can cause leaks. Contexts should be short-lived.

Alrighttt:
I removed all code related to the peer cancellation mechanism.

chain/db.go
```diff
@@ -65,6 +65,14 @@ type DB interface {
...
 	Flush() error
 	Cancel()
+
+	// View calls fn with a read-only DB. Implementations must guarantee
+	// that the DB passed to fn observes a consistent snapshot for the
+	// duration of fn, and that fn may run concurrently with other View
+	// calls. Buckets obtained via the DB inside fn must not be used
+	// after fn returns. Writes via the DB inside fn are not durable
+	// and may panic.
+	View(fn func(DB) error) error
```
lukechampine: this makes me sad

maybe we could add a `writeable` flag on `Bucket()` instead?

ChrisSchinnerl: Not sure you'd like that more but if you want to be able to do multiple operations within a single `View` we could also do something like 

```go
snap := m.store.Snapshot()
defer snap.Close()
for range 10 {
  _, _ = snap.Block(...)
}
```

Where `Snapshot` starts a readonly transaction which returns a `ReadonlyDB` interface only containing a subset of the `DB` interface. So it's impossible to call `CreateBucket` in the first place.

Alrighttt: Implemented Chris's suggestion

chain/db.go
```diff
@@ -948,6 +999,40 @@ func (db *DBStore) Flush() error {
...
+// so its inherited write methods are unreachable.
+type readonlyStore struct {
+	*DBStore
+	rdb ReadonlyDB
+}
+
+// Close releases the underlying snapshot.
+func (s *readonlyStore) Close() {
+	if err := s.rdb.Close(); err != nil {
+		panic(err)
```
ChrisSchinnerl: Let's not add more panics to this codebase. `Close` commonly returns an error (see `io.Closer` interface). The caller can always choose to ignore it.

ChrisSchinnerl (changes requested):

chain/db.go
Alrighttt: Removed in https://github.com/SiaFoundation/coreutils/pull/437/commits/94d628eccc29add6575554d2666e533a0d691b9a

ChrisSchinnerl (dismissed):

chain/db.go
lukechampine: I do like the idea of making it impossible to call mutating methods, but to make it airtight, we need a read-only `DBBucket` as well that has no `Put` or `Delete` methods.

Alrighttt: Added this in https://github.com/SiaFoundation/coreutils/pull/437/commits/38d9717ab9096366dcd62dba73de187dc3f64da3

ChrisSchinnerl (dismissed):

.changeset/change_chain_manager_mutex_to_rwmutex_to_enable_parallel_reads.md
```diff
@@ -0,0 +1,5 @@
+---
+default: patch
```
peterjan, line 2: This should probably be `minor` considering we're extending the exported `Store` and `DB` interfaces? Anyone implementing those will have to add Snapshot for it to compile.

chain/concurrency_test.go
```diff
@@ -0,0 +1,223 @@
...
+			return db
+		}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			runConcurrentReadsWrites(t, tc.makeDB(t))
+		})
+	}
+}
+
+func runConcurrentReadsWrites(t *testing.T, db chain.DB) {
```
peterjan, line 51: Should we add a reader that calls `AddV2PoolTransactions` alongside the existing readers? Isn't that the most interesting one because we temporarily release the lock there to notify the listeners?

chris124567 (commented):
LGTM besides PJ's comments

chain/concurrency_test.go
Alrighttt: Addressed in https://github.com/SiaFoundation/coreutils/pull/437/commits/1af204c9773303b00c08c638fea13de5a359e8e6

peterjan (approved):

ChrisSchinnerl (approved):

lukechampine:
this is not as clean as I'd hoped. I experimented with some alternative designs, but they didn't come out much better.

I think the issue is that `RWMutex` and BoltDB have different semantics. `RWMutex` doesn't give you a snapshot; it just blocks until the writer is finished. So even if we have this nice SnapshotDB that can serve requests even while the DB is being written to, we don't get to actually _use_ it, because it's gated behind an `RLock()`, which can't complete until `Unlock()` is called.

And yet, IIRC @Alrighttt said that, empirically, this change eliminated the issue. So I'm confused. I can understand how, if you have a bunch of concurrent readers, this design would no longer force them to be serialized. But that property would seem to break as soon as you have 1 writer in the mix (e.g. a malicious peer submitting txns to your pool, or drip-feeding you sidechain blocks).

Clearly, snapshot semantics are better than RWMutex, because you can continue serving requests even while writing. So maybe we should aim for snapshots of the whole Manager state? IIUC this would require making a deep copy of the transaction pool, but only when we start a write transaction, which doesn't seem too bad.

Alrighttt:
Closing in favor of https://github.com/SiaFoundation/coreutils/pull/450

PR #439, opened 2026-06-09
Add MaxInflightRPCsPerSubnet

Alrighttt:
Adds a per-subnet limit on concurrent inbound RPCs.


syncer/syncer.go
```diff
@@ -365,6 +441,7 @@ func (s *Syncer) runPeer(p *Peer) {
 	}
 	defer done()
 
+	subnet := s.subnetKey(p.ConnAddr)
 	inflight := make(chan struct{}, s.config.MaxInflightRPCs)
```
lukechampine, line 468: given that we already have a `MaxInflightRPCs` (per peer), might it make more sense to implement `MaxInflightRPCsPerSubnet` by adding `MaxPeersPerSubnet`?

lukechampine (dismissed):

syncer/syncer.go
```diff
@@ -136,6 +139,9 @@ func defaultConfig() config {
 		MaxInboundPeers:            64,
 		MaxOutboundPeers:           16,
 		MaxInflightRPCs:            64,
+		MaxInflightRPCsPerSubnet:   64,
+		InflightIPv4PrefixBits:     32,
+		InflightIPv6PrefixBits:     64,
```
ChrisSchinnerl: Isn't that quite high? e.g. my ISP gives me a /48 for free.

syncer/syncer.go
ChrisSchinnerl, line 1: Would be good to have testing for this

ChrisSchinnerl (changes requested):

syncer/syncer.go
Alrighttt: It's now set to /48

syncer/syncer.go
Alrighttt: Added CI tests to syncer_test.go

lukechampine (approved):

ChrisSchinnerl (approved):

PR #440, opened 2026-06-09
enforce reorg path length limit inside reorgPath

Alrighttt:
Adds a maxLen parameter to reorgPath and check it inside the rewind loop, so an over-long path short-circuits while it's being walked instead of being fully built and rejected afterward


lukechampine (dismissed):

chain/manager.go
```diff
@@ -432,9 +433,13 @@ func (m *Manager) applyTip(index types.ChainIndex) error {
 	return nil
 }
 
-func (m *Manager) reorgPath(a, b types.ChainIndex) (revert, apply []types.ChainIndex, err error) {
+func (m *Manager) reorgPath(a, b types.ChainIndex, maxLen int) (revert, apply []types.ChainIndex, err error) {
```
ChrisSchinnerl, line 436: It's usually good to have test coverage for anything related to validation. Especially in `core`.

ChrisSchinnerl (changes requested):

chain/manager.go
Alrighttt: Added some unit tests.

lukechampine (approved):

ChrisSchinnerl (approved):

PR #444, opened 2026-06-17
Remove txn from pool when contract formation or renewal fails

ChrisSchinnerl:
I ran into a bit of a loop with one of our hosts on Zeus today. The problem being a renewal failing due to an `internal error` and the likely candidate being the `RenewContract` call failing.
That causes the used outputs of the host to be released but the transaction remains in the transaction pool. So the renter can't reuse their outputs since the host's pool thinks they were double spent.
This leads to the following contract formation (triggered by the existing contract being stuck in the renew window) failing due to a double spent. Since the transaction is never broadcasted the indexer will never find out about that though.

This PR attempts to fix this by explicitly removing the transaction from the pool upon failure. Which should at least allow the indexer to form a new contract to replace the stuck one with.

### DIFF
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -1473,6 +1473,7 @@ func (m *Manager) RemoveV2PoolTransactions(ids []types.TransactionID) {
 	// force a full revalidation
 	m.txpool.ms = nil
 	m.txpool.medianFee = nil
+	m.txpool.weight = 0
 	m.revalidatePool()
 
 	// release lock while notifying listeners
--- a/chain/manager.go
+++ b/chain/manager.go
@@ -1443,6 +1443,50 @@ func (m *Manager) AddV2PoolTransactions(basis types.ChainIndex, txns []types.V2T
 	return false, nil
 }
 
+// RemoveV2PoolTransactions removes the given transactions from the transaction
+// pool, along with any pooled transactions that depend on them. It is intended
+// for abandoning a locally-added transaction set that will not be broadcast.
+// Transactions that are not in the pool are ignored.
+func (m *Manager) RemoveV2PoolTransactions(ids []types.TransactionID) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	remove := make(map[types.TransactionID]bool, len(ids))
+	for _, id := range ids {
+		remove[id] = true
+	}
+
+	filtered := m.txpool.v2txns[:0]
+	var removed bool
+	for _, txn := range m.txpool.v2txns {
+		if remove[txn.ID()] {
+			removed = true
+			continue
+		}
+		filtered = append(filtered, txn)
+	}
+	if !removed {
+		return
+	}
+	m.txpool.v2txns = filtered
+
+	// force a full revalidation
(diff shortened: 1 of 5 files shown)


chain/manager.go
```diff
@@ -1443,6 +1443,51 @@ func (m *Manager) AddV2PoolTransactions(basis types.ChainIndex, txns []types.V2T
 	return false, nil
 }
 
+// RemoveV2PoolTransactions removes the given transactions from the transaction
+// pool, along with any pooled transactions that depend on them. It is intended
+// for abandoning a locally-added transaction set that will not be broadcast.
+// Transactions that are not in the pool are ignored.
+func (m *Manager) RemoveV2PoolTransactions(ids []types.TransactionID) {
```
n8mgr, line 1553: This is quite dangerous.. Not sure I like it. There’s no way to tell if the transaction has been broadcast so it’d be easy to create harder to debug issues

ChrisSchinnerl: What if we add an optional arg to `BroadcastV2TransactionSet` to prevent persistence of the transaction iff the broadcast failed and a `ErrBroadcastFailed` that we can check for.
To make sure that 1. we didn't broadcast and 2. we won't retry.

Because right now we have a similar issue on `master`. `BroadcastV2TransactionSet` might fail, we release the inputs but then we might retry later and double spend.

n8mgr: I think we'd need to go a lot further to get the correct fix instead of a somewhat dangerous patch. Needs a lot of thought on potential edge cases since we don't have anything like replace by fee in place.

We started persisting broadcasted txns to prevent double spend attempts during restarts in addition to retrying if the txn doesn't make it into the next block. 

We effectively need to check if the txn is valid for our tpool and attempt a broadcast before actually adding it to our local pool and persisting it preferably under lock, but that comes with its own edge cases. If broadcast fails, we realistically don't want it in our local tpool locking resources either but that's also the safest option and it's temporary. 

The smallest change would be to reverse the order, but it doesn't fully solve the issue.
```go
if err := syncer.Broadcast() {

} else if err := store.AddBroadcastedSet() {

}
```

ChrisSchinnerl: Yeah I considered changing the order but I wasn't sure if we wanted to change the behavior of "if it's in the pool we definitely rebroadcast". 

But I just realized. Doesn't `hostd` support reverting a renewal now when a renewed contract never makes it on chain? Thanks to the CoW like behavior you added for faster renewals? Because that makes me wonder if the following approach is now possible:

In `handleRPCRenewContract` instead of calling `s.chain.AddV2PoolTransaction` we call a new `s.chain.ValidateV2PoolTransaction` optimistically. Then we call `s.contractor.RenewV2Contract` followed by `s.wallet.BroadcastV2TransactionSet`. 

There is a gap between validating the txn set before the renewal and actually adding it in the broadcasting call but maybe that is alright now? If validation fails while broadcasting, we end up with a contract in our database that has definitely not been broadcasted or added to the pool. So it should expire. The error the client gets is then "contract is already renewed" for a few hours before it then succeeds on the next try hopefully.

PR #450, opened 2026-07-04
chain: Overhaul persistence layer

lukechampine:
This supersedes #437, implementing true MVCC snapshots for the Manager. Read-only methods (like `Tip()`, `BlocksForHistory()`, etc.) see a consistent "snapshot" of the Manager state, while state-mutating methods (like `AddBlocks`) operate on a "scratchpad," that, once committed, becomes the next snapshot. These are equivalent to BoltDB's `View` and `Update`, respectively.

The effect of this is that readers can't block writers, and writers can't block readers. Writers do block other writers -- for example, `AddBlocks` calls still need to be serialized -- but that's it.

This code is a pretty even split between myself and Claude. I designed the interfaces and refined the APIs to my liking, while Claude handled the mechanical refactoring, the MemDB updates, and writing tests. I suspect there is some slop lurking in here that will turn up in review, but overall I'm happy with the result.

### DIFF
--- a/.changeset/support_concurrent_chain_manager_reads_via_mvcc_snapshots.md
+++ b/.changeset/support_concurrent_chain_manager_reads_via_mvcc_snapshots.md
@@ -1,5 +1,5 @@
 ---
-default: minor
+default: major
 ---
 
 # Support concurrent chain.Manager reads via MVCC snapshots
--- a/.changeset/support_concurrent_chain_manager_reads_via_mvcc_snapshots.md
+++ b/.changeset/support_concurrent_chain_manager_reads_via_mvcc_snapshots.md
@@ -1,0 +1,14 @@
+---
+default: minor
+---
+
+# Support concurrent chain.Manager reads via MVCC snapshots
+
+The Manager previously serialized all of its methods -- including read-only ones -- with a single mutex, so heavy read load (e.g. from the syncer or subscribers) contended with block processing. The Manager now follows an MVCC scheme: writes accumulate in the Store's "scratchpad" -- which includes the tip state -- and are committed by Flush, at which point they become visible to "snapshots": self-contained, read-only captures of the committed state. Read-only Manager methods operate on snapshots, so they proceed concurrently with each other and with block processing, and never stall writers. bbolt supports snapshots natively via read-only transactions; `MemDB` now implements them with pin-aware copy-on-flush generations (`CacheDB` delegates to its underlying DB).
+
+This is a breaking change for implementors and consumers of the `chain.DB` and `chain.Store` interfaces:
+
+- `Store` is now `Snapshot() (StoreSnapshot, func())` + `Scratchpad() StoreScratchpad`. `StoreSnapshot` carries the read methods plus `TipState`; `StoreScratchpad` embeds it and adds the write methods and `Flush`, with `ApplyBlock`/`RevertBlock` updating the scratchpad's tip. Implementations may flush autonomously between operations (e.g. to bound the size or age of the accumulated writes), handling errors internally; the Manager explicitly calls Flush to publish the new tip after processing blocks, before notifying OnReorg listeners.
+- `DB` mirrors `Store`: `Snapshot() (DBSnapshot, release func())` + `Scratchpad() DBScratchpad`. The scratchpad carries `Bucket`/`CreateBucket`/`Flush`/`Cancel` and must not flush autonomously; snapshots are read-only and remain valid until released, even across concurrent Flushes. A single `DBBucket` interface serves both sides; its `Put`/`Delete` methods may return errors on read-only (snapshot) buckets.
+- `NewManager` no longer takes a `consensus.State` parameter, and `NewDBStore`/`NewDBStoreAtCheckpoint` no longer return one; the tip state is provided by the Store itself (via a snapshot's or scratchpad's `TipState`).
+- Reads reflect committed state. Pruned blocks remain visible to read-only methods until the next commit, and blocks added to non-best chains remain *invisible* to them until the next commit. Blocks that extend the best chain are always committed before `AddBlocks` returns and before `OnReorg` listeners fire, so subscribers observe the tip they are notified of.
--- a/chain/db.go
+++ b/chain/db.go
@@ -6,8 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"iter"
+	"maps"
 	"math/bits"
 	"sort"
+	"sync"
 	"time"
 
 	"go.sia.tech/core/consensus"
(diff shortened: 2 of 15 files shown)


chain/db.go
```diff
@@ -178,7 +283,7 @@ func (b memBucket) Put(key, value []byte) error { return b.db.put(b.name, key, v
 func (b memBucket) Delete(key []byte) error     { return b.db.delete(b.name, key) }
 func (b memBucket) Iter() iter.Seq2[[]byte, []byte] {
 	return func(yield func([]byte, []byte) bool) {
-		for key, val := range b.db.buckets[b.name] {
+		for key, val := range b.db.gen.buckets[b.name] {
 			if pval, ok := b.db.puts[b.name][string(key)]; ok {
 				val = pval
```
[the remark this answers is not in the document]
lukechampine: (pre-existing, and never reached in production code, but fixed anyway)

chain/db.go
```diff
@@ -861,121 +957,268 @@ func (db *DBStore) AncestorTimestamp(id types.BlockID) (t time.Time, ok bool) {
...
+
+// ApplyBlock implements StoreScratchpad.
+func (s *dbScratchpad) ApplyBlock(cs consensus.State, cau consensus.ApplyUpdate) {
+	s.bucket(bMainChain).put(encHeight(cs.Index.Height), &cs.Index.ID)
+	s.bucket(bMainChain).putRaw(keyHeight, encHeight(cs.Index.Height))
+	if cs.Index.Height <= s.n.HardforkV2.RequireHeight {
+		s.applyElements(cau)
+	}
+	s.tip = cs
+	if s.shouldFlush() {
```
Alrighttt, line 1091: Why not use `flushIfFull` in this case? A snapshot reader could briefly see the tip height go down whereas previously tip height was monotonic

chain/db_test.go
```diff
@@ -5,52 +5,21 @@ import (
...
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils"
 	"go.sia.tech/coreutils/chain"
 	"go.sia.tech/coreutils/testutil"
 	"lukechampine.com/frand"
 )
 
-// NOTE: due to a bug in the transaction validation code, calculating payouts
```
Alrighttt, line 19: Removing this comment seems unrelated to this PR?

Alrighttt (dismissed):
LGTM; I prefer this over https://github.com/SiaFoundation/coreutils/pull/437 .

.changeset/support_concurrent_chain_manager_reads_via_mvcc_snapshots.md
```diff
@@ -0,0 +1,14 @@
+---
+default: minor
```
ChrisSchinnerl: Probably deserves a `major` since it breaks every single piece of node software we got. e.g. `chain.NewManager` also has a different signature now without the `tipState`.

chain/db.go
```diff
@@ -59,8 +61,32 @@ func (vs *versionedState) DecodeFrom(d *types.Decoder) {
...
+// A DBSnapshot is a read-only snapshot of a DB.
+type DBSnapshot interface {
+	Bucket(name []byte) DBBucket
+}
+
+// A DBScratchpad accumulates writes to a DB, which become durable when Flush
+// is called. Its methods observe the accumulated writes. It is not safe for
+// concurrent use.
+//
+// Unlike a StoreScratchpad, a DBScratchpad must not flush autonomously:
```
ChrisSchinnerl, line 86: (nit) slightly confusing that a `DBScratchpad` must not flush autonomously but a `dbScratchpad` flushes all the time because it's not actually a `DBScratchpad` but a `StoreScratchpad`

chain/db.go
ChrisSchinnerl: Deletions also contribute to `unflushed` so I think using `flushIfFull` isn't enough here. Either we explicitly skip flushing when the resulting height is lower or we prevent `delete` from incrementing `unflushed`.

But also this just makes it feel like `dbScratchpad` is too low level to be able to decide whether flushing is safe.

chain/manager_test.go
```diff
@@ -674,3 +684,172 @@ func TestReorgPathBogusBasisBailsFast(t *testing.T) {
...
+		t.Fatal("expected tip height 25, got", cm.Tip().Height)
+	}
+}
+
+// TestSnapshotDoesNotBlockWrites verifies that a long-lived snapshot neither
+// stalls writers nor observes their effects.
+func TestSnapshotDoesNotBlockWrites(t *testing.T) {
+	n, genesisBlock := TestnetZen()
+	genesisBlock.Transactions = nil
+	n.InitialTarget = types.BlockID{0xFF}
```
peterjan, line 914: `nit` could use a helper like this to get rid of a bunch of setup 

```
// testV2Network returns the Zen testnet with a trivial difficulty target and
// all hardforks active from height 1.
func testV2Network() (*consensus.Network, types.Block) {
       n, genesisBlock := TestnetZen()
       n.InitialTarget = types.BlockID{0xFF}
       n.BlockInterval = time.Second
       n.MaturityDelay = 5
       n.HardforkDevAddr.Height = 1
       n.HardforkTax.Height = 1
       n.HardforkStorageProof.Height = 1
       n.HardforkOak.Height = 1
       n.HardforkASIC.Height = 1
       n.HardforkFoundation.Height = 1
       n.HardforkV2.AllowHeight = 1
       n.HardforkV2.RequireHeight = 1
       n.HardforkV2.FinalCutHeight = 1
       return n, genesisBlock
}
```

chain/db.go
```diff
@@ -861,121 +957,268 @@ func (db *DBStore) AncestorTimestamp(id types.BlockID) (t time.Time, ok bool) {
...
+	return getBlockHeader(s.ss, id)
+}
+
+// State implements StoreSnapshot.
+func (s *dbSnapshot) State(id types.BlockID) (consensus.State, bool) {
+	return getState(s.ss, s.n, id)
+}
+
+// ExpiringFileContractIDs implements StoreSnapshot.
+func (s *dbSnapshot) ExpiringFileContractIDs(height uint64) []types.FileContractID {
```
peterjan, line 1167: these read methods are shared between `dbSnapshot` and `dbScratchpad`, there's quite a few so perhaps it makes sense to clean that up a bit by adding a small `dbReader` struct that implements them

chain/manager.go
```diff
@@ -600,13 +644,13 @@ func (m *Manager) UpdatesSince(index types.ChainIndex, maxBlocks int) (rus []Rev
 // chain changes. The fn is called with the new tip. It returns a function that
 // removes fn from the set.
 func (m *Manager) OnReorg(fn func(types.ChainIndex)) (cancel func()) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	_, unlock := m.scratchpad()
```
peterjan, line 647: why not acquire the lock directly, we're discarding the scratchpad anyway

chain/manager.go
```diff
@@ -615,20 +659,21 @@ func (m *Manager) OnReorg(fn func(types.ChainIndex)) (cancel func()) {
 // transaction pool may have changed. It returns a function that removes fn from
 // the set.
 func (m *Manager) OnPoolChange(fn func()) (cancel func()) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	_, unlock := m.scratchpad()
```
peterjan, line 662: why not acquire the lock directly, we're discarding the scratchpad anyway

chain/manager_test.go
```diff
@@ -674,3 +684,172 @@ func TestReorgPathBogusBasisBailsFast(t *testing.T) {
...
+	store, err := NewDBStore(NewMemDB(), n, genesisBlock, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cm := NewManager(store)
+
+	var wg sync.WaitGroup
+	stop := make(chan struct{})
+	for range 8 {
+		wg.Add(1)
```
peterjan: `nit` use wg.Go

chain/manager.go
```diff
@@ -1407,9 +1452,9 @@ func (m *Manager) UpdateV2TransactionSet(txns []types.V2Transaction, from, to ty
 	if from == to {
 		return txns, nil
 	}
-	m.mu.Lock()
-	defer m.mu.Unlock()
-	return m.updateV2TransactionProofs(txns, from, to)
+	sp, unlock := m.scratchpad()
```
peterjan: I think technically this could be a snapshot (?) - but probably safer to keep this. Just figured I'd flag it because I don't fully understand how big of a win it would be if we can use the snapshot as opposed to the scratch pad

peterjan (commented):
couple of minor comments but LGTM over all

chain/db_test.go
lukechampine: it is indeed irrelevant

chain/manager.go
lukechampine: for consistency. "Every method calls either `snapshot` or `scratchpad`" is easy to spot-check.

chain/manager.go
lukechampine: good catch

chain/db.go
lukechampine: hmm yeah those names are not ideal lol

lukechampine:
@copilot resolve the merge conflicts in this pull request

Alrighttt (approved):

